PR #1, opened 2019-03-01
Added Tracker to window.paypal

bryanjenningz:


__sdk__.js
```diff
@@ -6,5 +6,10 @@ module.exports = {
         entry:           './src/index',
         staticNamespace: '__muse__',
         automatic:        true
+    },
+    'tracker': {
```
shivinkapur: is there any naming convention to be followed here?

src/tracker.js
```diff
@@ -0,0 +1,3 @@
+/* @flow */
+
+export { Tracker } from './tracker-component';
```
shivinkapur: how can one be sure Tracker object is available on the global?

src/tracker-component.js
```diff
@@ -0,0 +1,40 @@
+/* @flow */
+
+import { getClientID } from '@paypal/sdk-client/src';
```
shivinkapur: can `getClientId` be undefined?

src/tracker.js
joshbeam: it's magic. when components get pulled into the sdk, any exports are automatically attached to window.paypal

__sdk__.js
```diff
@@ -6,5 +6,10 @@ module.exports = {
...
+        entry:           './src/tracker',
+        staticNamespace: '__tracker__',
+        automatic:        true
```
bluepnume: Can we remove `automatic: true`? This is a new integration pattern, so merchants should pass `components=buttons,tracker` when integrating this. That way they'll only get the specific code they need.

src/tracker-component.js
joshbeam: the result of it? yes, but then the sdk won't work at all. `getClientId` itself though, no won't be undefined

bluepnume: Nope.

src/tracker-component.js
```diff
@@ -0,0 +1,40 @@
...
+        img.addEventListener('load', resolve);
+        img.addEventListener('error', reject);
+        document.body.appendChild(img);
+    });
+};
+
+const trackCartEvent = (cartEventType, trackingData) =>
+    track('cartEvent', { ...trackingData, cartEventType });
+
+export const Tracker = {
```
bluepnume: Can we call this `CartTracker` maybe so it's obvious what the namespace represents? Given the number of other trackers we're surfacing in this sdk.

src/tracker.js
bluepnume: 👍

src/tracker-component.js
shivinkapur: which are the other trackers on the sdk? Any easy way to know that?

bluepnume:
Agreed on the tests :) we're putting a lot of eggs in one basket with the sdk, so need to be very confident for each release that everything is working the way it should

src/tracker-component.js
bluepnume: Well muse already does tracking, we have fpti tracking, cal tracking, etc. Just want to make sure any public `paypal.*` names aren't too ambiguous before we set them in stone.

bluepnume: Would it make sense to do something like:

```
let cartTracker = paypal.CartTracker({ userID, userName })
cartTracker.addToCart();
```

That would avoid the need to use a global `user`, and also makes the module less of a singleton / potentially extendable in future if there are any partners or marketplaces who have multiple carts, or anything like that.

On that node -- seems like it would be a good idea to accept a `cartID` in that constructor here so we can disambiguate multiple carts in future, especially for partner cases.

src/tracker-component.js
```diff
@@ -0,0 +1,40 @@
...
+    setUser: data => {
+        user.id = data.user.id;
+        user.name = data.user.name;
+    },
+    view:           track.bind(0, 'view'),
```
bluepnume: Can we avoid doing bind at the top-level module scope? This will be counted as a side effect, which will kick webpack/uglify into non tree shaking mode and increase the size of the whole bundle. That could also be solved by turning `paypal.CartTracker` into a function as above.

__sdk__.js
bluepnume: I'd suggest we use `cart-tracker` here too, since we'd be expecting merchants to pass this explicitly in the sdk url without `automatic: true`.

joshbeam: I think the tracker was gonna end up being used for more than just cart stuff, though. Correct me if I'm wrong @shivinkapur @bryanjenningz

src/tracker-component.js
joshbeam: :+1: I second the above style: `let cartTracker = paypal.CartTracker({ userID, userName })`

__sdk__.js
bluepnume: OK. If that's the case then makes sense. 👍

src/tracker-component.js
arvindrajnaidu: The idea is to track all events on a merchant's page, not just cart events. It really is the FPTI client for merchants to implement. I am working on the docs here https://github.paypal.com/arnaidu/tracker/blob/master/PayPalShopping.md

I feel we should not mandate 3 different objects ... PageTracker, CartTracker, PurchaseTracker in order to track events. Instead we were thinking one instance Tracker that one uses like this. Tracker.addToCart, Tracker.purchase, Tracker.pageView.

```

Comment (arvindrajnaidu) [reply src/tracker-component.js]: All events will have to carry the clientId and merchantId if available. This will solve for the partner and market places no?

Comment (joshbeam) [reply src/tracker-component.js]: @arvindrajnaidu I agree with you. I was second'ing the "let's not set a global user" thing.

Comment (bluepnume) [reply src/tracker-component.js]: That's fair. If we're going to use this for multiple use-cases let's go with `paypal.Tracker`. But let's make it a constuctor (`let tracker = paypal.Tracker({ userID, userName })`)

File: test/tracker-component.js [hunk 7]
```diff
@@ -0,0 +1,148 @@
...
+                    id:  '__test__productId',
+                    url: 'https://example.com/__test__productId'
+                }
+            ],
+            emailCampaignId: '__test__emailCampaignId',
+            price:           12345.67,
+            currencyCode:    'USD',
+            keywords:        [ '__test__' ]
+        });
+        expect(appendChildCalls).to.equal(2);
```
joshbeam: Kind of an anti-pattern here for unit tests to depend on the state of the previous tests (i.e. instead of just incrementing the `appendChildCalls` for each test, increment it once per test and add an `afterEach` that resets it back to its original state). As it is, this particular test suite is kind of brittle, because if you were to change one of these tests or remove it, then all the other tests would break. This is a moot point though if you instead change your assertions to check the img.src instead of just checking that `appendChild` was called.

test/tracker-component.js
```diff
@@ -0,0 +1,148 @@
+/* globals describe before it */
```
joshbeam: I'd add a test in here somewhere that asserts that you can instantiate trackers with different users

src/tracker-component.js
```diff
@@ -0,0 +1,102 @@
...
+        const data = {
+            ...trackingData,
+            user:       config.user,
+            property:   config.property,
+            trackingType,
+            clientId:   getClientID(),
+            merchantId: getMerchantID()
+        };
+
+        if (config.paramsToBeaconUrl) {
```
joshbeam: Can you add a comment here that describes the purpose of paramsToBeaconUrl?

src/tracker-component.js
```diff
@@ -0,0 +1,102 @@
...
+const trackCartEvent = <T>(config : Config, cartEventType : CartEventType, trackingData : T) : Promise<void> =>
+    track(config, 'cartEvent', { ...trackingData, cartEventType });
+
+export const Tracker = (config : Config) => ({
+    view:           (data : { pageUrl : string }) => track(config, 'view', data),
+    addToCart:      (data : Cart) => trackCartEvent(config, 'addToCart', data),
+    setCart:        (data : Cart) => trackCartEvent(config, 'setCart', data),
+    removeFromCart: (data : RemoveCart) => trackCartEvent(config, 'removeFromCart', data),
+    purchase:       (data : { cartId : string }) => track(config, 'purchase', data),
+    setUser:        (data : { user : { id : string, name? : string, email? : string } }) => {
```
joshbeam: why do we have this method `setUser`?

test/tracker-component.js
```diff
@@ -0,0 +1,148 @@
...
+    it('should be a function that returns a tracker', () => {
+        const userID = '__test__userID';
+        const userName = '__test__userName';
+        const tracker = Tracker({ user: { id: userID, name: userName } });
+        expect(tracker).to.have.property('view');
+        expect(tracker).to.have.property('addToCart');
+        expect(tracker).to.have.property('setCart');
+        expect(tracker).to.have.property('removeFromCart');
+        expect(tracker).to.have.property('purchase');
+        expect(tracker).to.not.have.property('purchases');
```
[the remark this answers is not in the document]
bryanjenningz: Oh, I just wanted to make the assertion fail to make sure it was working, I forgot to remove this. I'll remove it now.

src/tracker-component.js
bryanjenningz: It's possible that we don't know the user when we start tracking their cart events, but later they might log in and we'll know who they are, so we can use `setUser`.

test/tracker-component.js
```diff
@@ -0,0 +1,148 @@
...
+
+    it('should send view events', () => {
+        const userID = '__test__userID2';
+        const userName = '__test__userName2';
+        const tracker = Tracker({ user: { id: userID, name: userName } });
+        expect(appendChildCalls).to.equal(0);
+        tracker.view({
+            pageUrl: 'https://example.com/test2'
+        });
+        expect(appendChildCalls).to.equal(1);
```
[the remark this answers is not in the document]
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

test/tracker-component.js
```diff
@@ -0,0 +1,305 @@
...
+            JSON.stringify({
+                cartId: '__test__cartId',
+                items:  [
+                    {
+                        id:  '__test__productId',
+                        url: 'https://example.com/__test__productId'
+                    }
+                ],
+                emailCampaignId: '__test__emailCampaignId',
+                price:           12345.67,
```
joshbeam: @bryanjenningz One minor thing regarding numbers: do you want to accept floats here? (Typical problems with JavaScript floating point numbers). I'm not sure what kind of backend processing happens when it receives this data, but my initial reaction is that weird bugs could be encountered in the future.

The Payments SDK, for example, uses [strings](https://developer.paypal.com/docs/checkout/integrate/#5-capture-the-transaction) (e.g. `'12345.67'`), and another strategy for this is [zero decimal currency](https://stripe.com/docs/currencies#zero-decimal).

@bluepnume what convention do you guys think is best, cents as strings?

Probably worth deciding on this before merging.

test/tracker-component.js
```diff
@@ -0,0 +1,148 @@
...
+    let appendChildCalls = 0;
+    const appendChild = () => {
+        appendChildCalls += 1;
+    };
+    before(() => {
+        // $FlowFixMe
+        document.body = document.body || {};
+        // $FlowFixMe
+        document.body.appendChild = appendChild;
+    });
```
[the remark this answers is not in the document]
bryanjenningz: Done

test/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+    description? : string,
+    imgUrl? : string,
+    otherImages? : $ReadOnlyArray<string>
+|};
+
+type Cart = {|
+    cartId? : string,
+    items : $ReadOnlyArray<Product>,
+    emailCampaignId? : string,
+    price? : number,
```
bryanjenningz: total? : string

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+type TrackingType = 'view' | 'cartEvent' | 'purchase';
+
+type CartEventType = 'addToCart' | 'setCart' | 'removeFromCart';
+
+type Environment = 'stage' | 'sandbox' | 'production';
+
+type Product = {|
+    id : string,
+    url : string,
+    sku? : string,
```
bryanjenningz: remove sku, we only care about id for being able to identify the product

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+    name? : string,
```
bryanjenningz: title? : string instead of name
if the user has a product that doesn't have a title, the product has to have been ingested

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+    description? : string,
+    imgUrl? : string,
```
bryanjenningz: rename imageUrl

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+
+type TrackingType = 'view' | 'cartEvent' | 'purchase';
+
+type CartEventType = 'addToCart' | 'setCart' | 'removeFromCart';
+
+type Environment = 'stage' | 'sandbox' | 'production';
+
+type Product = {|
+    id : string,
+    url : string,
```
bryanjenningz: url? : string

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+        id : string
+    },
+    paramsToBeaconUrl? : ParamsToBeaconUrl,
+    env? : Environment
+|};
+
+const track = <T>(config : Config, trackingType : TrackingType, trackingData : T) : Promise<void> => {
+    const encodeData = data => encodeURIComponent(btoa(JSON.stringify(data)));
+
+    return new Promise((resolve, reject) => {
```
bluepnume: This is going to break in older browsers. Would recommend using `import { ZalgoPromise } from 'zalgo-promise/src'` which we use in other libs in the sdk.

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+    user : {
+        id : string,
+        email? : string, // mandatory if unbranded cart recovery
+        name? : string
+    },
+    property? : {
+        id : string
+    },
+    paramsToBeaconUrl? : ParamsToBeaconUrl,
+    env? : Environment
```
bryanjenningz: remove env

bluepnume:
This is looking good to me 👍 no major concerns.

src/tracker-component.js
```diff
@@ -0,0 +1,105 @@
...
+    items : $ReadOnlyArray<{ id : string }>
+|};
+
+type ParamsToBeaconUrl = ({
+    trackingType : TrackingType,
+    data : Cart | RemoveCart | { cartId : string } | { pageUrl : string }
+}) => string;
+
+type Config = {|
+    user : {
```
bryanjenningz: optional user object

test/tracker-component.js
bryanjenningz: Fixed this, it uses '12345.67' format now

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done, removed promises, it's just fire and forget now

src/tracker-component.js
bryanjenningz: Done

src/tracker-component.js
bryanjenningz: Done

PR #2, opened 2019-03-15
Add messaging component to muse components 

zacharyhill:
Create exit modal for collecting email to send email with products

PR includes previous PR https://github.com/paypal/paypal-muse-components/pull/1


src/messaging-component.js
```diff
@@ -0,0 +1,115 @@
...
+import { loadJavascript } from './lib/load-js';
+
+const museSdkUrl = 'https://ppsong.c0d3.com/cdn/assets/modal/sdk.js';
+loadJavascript(museSdkUrl);
+
+let userId = '';
+let isRendered = false;
+
+// $FlowFixMe
+ppq('showExperience', 'https://www.paypalobjects.com/muse/cart-recovery-0.3/', 'body', {
```
bluepnume: We should avoid kicking off side-effects in the module scope. It makes testing tricky, and also de-optimizes any tree shaking that can be done by webpack.

The SDK allows you to export a `setup` function which runs on load. Can we use that instead? For example: https://github.com/paypal/paypal-checkout-components/blob/master/src/interface/button.js#L51

src/messaging-component.js
```diff
@@ -0,0 +1,115 @@
+/* globals ppq */
+/* @flow */
+
+import { Tracker } from './tracker-component';
+import { checkIfMobile } from './lib/mobile-check';
+import { loadJavascript } from './lib/load-js';
+
+const museSdkUrl = 'https://ppsong.c0d3.com/cdn/assets/modal/sdk.js';
+loadJavascript(museSdkUrl);
```
bluepnume: Same here. Let's avoid side effects in the module scope.

src/tracker-component.js
```diff
@@ -98,24 +123,22 @@ const track = <T>(config : Config, trackingType : TrackingType, trackingData : T
...
-    Math.random()
-        .toString(16)
-        .slice(2);
-
-export const Tracker = (config? : Config = { user: { id: generateId() } }) => ({
+export const Tracker = (config? : Config = { user: { email: undefined, name: undefined } }) => ({
     view:           (data : ViewData) => track(config, 'view', data),
-    addToCart:      (data : CartData) => trackCartEvent(config, 'addToCart', data),
+    addToCart:      (data : CartData) => {
+        localStorage.setItem('paypal-cr-cart', JSON.stringify(data));
```
bluepnume: We should null-check localStorage.

src/messaging.js
```diff
@@ -0,0 +1,2 @@
+/* @flow */
+export { Messaging } from './messaging-component';
```
bluepnume: You will need to export `setup` here too

src/messaging-component.js
```diff
@@ -0,0 +1,116 @@
...
+    // $FlowFixMe
+    ppq('showExperience', 'https://www.paypalobjects.com/muse/cart-recovery-0.3/', 'body', {
+        sessionId:          'BOND007',
+        variant:            'modal',
+        flow:               'cart-recovery',
+        mobileVariant:      'modal',
+        mobileFlow:         'cart-recovery',
+        isMobileEnabled:    'true',
+        isDesktopEnabled:   'true',
+        mrid:               'FY8HBGU6Y2MWW',
```
bluepnume: Do we want to hard-code this and other params here?

src/messaging-component.js
```diff
@@ -0,0 +1,116 @@
...
+            command:    'SHOW_OVERLAY',
+            visitorId:  userId
+        });
+        setCookie('paypal-cr-lastseen', 'true', sevenDays);
+        isRendered = true;
+    }
+    return isRendered;
+};
+
+const init = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string } }>) => {
```
bluepnume: For the `Tracker` component y'all went with a constructor pattern: https://github.com/paypal/paypal-muse-components/pull/1 (rather than having an init which modifies globals). For consistency should we do the same here?

```
const messaging = Messaging({
    userID: 'foo'
});
```

src/messaging-component.js
```diff
@@ -0,0 +1,116 @@
...
+};
+
+export const Messaging = {
+    init
+};
+
+export function setup() {
+    loadJavascript(museSdkUrl);
+    // $FlowFixMe
+    ppq('showExperience', 'https://www.paypalobjects.com/muse/cart-recovery-0.3/', 'body', {
```
bluepnume: Is this a race-condition? Isn't `loadJavascript` asynchronous?

src/lib/load-js.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
+
+export const loadJavascript = (url : string) => {
```
bluepnume: Would be good if we could make this non-minified and just have the compiler take care of this. Difficult to see exactly what this function is doing otherwise.

src/messaging-component.js
songz: currently, this is what the modal sdk from muse expects and there's nothing on the roadmap that requires us to change any of the fields, so we do want to hardcode these values in.

src/messaging-component.js
songz: loadJovascript sets up gloabal (ppq) and loads the arguments in. When the script is loaded, the script will run the appropriate function.

src/messaging-component.js
shivinkapur: @songz don't believe we need mrid here...similarly sessionId. Lets connect on these params

src/messaging-component.js
songz: currently, there's no user actions that are available for messaging component, its just set and go. We could return and empty object to be consistent but I don't think it accomplishes anything. what do you think @bluepnume ?

src/lib/load-js.js
songz: done

src/messaging-component.js
bluepnume: I don't mind if `Messaging` returns null -- mainly thinking about consistency with setting it up and passing in options. Right now:

Tracker -> `Tracker({ ...opts })`
Messaging -> `Messaging.init({ ...opts })`

PR #4, opened 2019-03-21
Use cookie for id, don't allow user to pass in id

bryanjenningz:
This makes it so the user never passes in an id, we generate it for them in all cases.

We get the id now from one of the following in order:
1. grab the id from document.cookie if it exists
2. if there's no id in document.cookie, we check localStorage
3. if there's no id in localStorage, the id gets generated and is put in localStorage then used as the id

### DIFF
--- a/src/cookie-utils.js
+++ b/src/cookie-utils.js
@@ -1,23 +1,23 @@
 /* @flow */
 
-export const getCookie = (cname : string) : string => {
-    const name = `${ cname }=`;
-    const ca = document.cookie.split(';');
-    for (let i = 0; i < ca.length; i++) {
-        let c = ca[i];
-        while (c.charAt(0) === ' ') {
-            c = c.substring(1);
+export const getCookie = (cookieName : string) : string => {
+    const name = `${ cookieName }=`;
+    const cookies = document.cookie.split(';');
+    for (let i = 0; i < cookies.length; i++) {
+        let cookie = cookies[i];
+        while (cookie.charAt(0) === ' ') {
+            cookie = cookie.slice(1);
         }
-        if (c.indexOf(name) === 0) {
-            return c.substring(name.length, c.length);
+        if (cookie.indexOf(name) === 0) {
+            return cookie.slice(name.length, name.length + cookie.length);
         }
     }
     return '';
 };
 
-export const setCookie = (cname : string, cvalue : string, exMilliseconds : number) : void => {
+export const setCookie = (cookieName : string, cookieValue : string, expirationMilliseconds : number) : void => {
     const d = new Date();
-    d.setTime(d.getTime() + exMilliseconds);
+    d.setTime(d.getTime() + expirationMilliseconds);
     const expires = `expires=${ d.toUTCString() }`;
-    document.cookie = `${ cname }=${ cvalue }; Path=/; ${ expires }`;
+    document.cookie = `${ cookieName }=${ cookieValue }; Path=/; ${ expires }`;
 };
--- a/src/cookie-utils.js
+++ b/src/cookie-utils.js
@@ -1,0 +1,23 @@
+/* @flow */
+
+export const getCookie = (cname : string) : string => {
+    const name = `${ cname }=`;
+    const ca = document.cookie.split(';');
+    for (let i = 0; i < ca.length; i++) {
+        let c = ca[i];
+        while (c.charAt(0) === ' ') {
+            c = c.substring(1);
+        }
+        if (c.indexOf(name) === 0) {
+            return c.substring(name.length, c.length);
+        }
+    }
+    return '';
+};
+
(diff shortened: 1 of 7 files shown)


src/tracker-component.js
```diff
@@ -66,15 +66,32 @@ type Config = {|
     paramsToBeaconUrl? : ParamsToBeaconUrl
 |};
 
+const getUserId = () => {
+    return document.cookie ? document.cookie : localStorage.getItem('user-id');
```
zacharyhill: should we make the localStorage key something that is less likely to be set by someone else?

zacharyhill: Also why are we returning all cookies?

src/tracker-component.js
```diff
@@ -66,15 +66,40 @@ type Config = {|
...
+    }
+    return userCookie.split('=')[1];
+};
+
+const getUserId = () : ?string => {
+    return getUserIdCookie() || localStorage.getItem('user-id');
+};
+
+const setRandomUserId = () : void => {
+    localStorage.setItem('user-id', generateId());
```
songz: should we prefix it with paypal? `paypal-user-id`?

src/tracker-component.js
bluepnume: Please null-check `localStorage`. It's not available all the time -- for some browsers, in private mode, it will not be defined.

src/tracker-component.js
```diff
@@ -66,15 +66,32 @@ type Config = {|
     paramsToBeaconUrl? : ParamsToBeaconUrl
 |};
 
+const getUserId = () => {
+    return document.cookie ? document.cookie : localStorage.getItem('user-id');
+};
+
+const setRandomUserId = () => {
+    localStorage.setItem('user-id', generateId());
```
bluepnume: Same here.

src/tracker-component.js
```diff
@@ -98,24 +115,19 @@ const track = <T>(config : Config, trackingType : TrackingType, trackingData : T
...
     removeFromCart: (data : RemoveCartData) => trackCartEvent(config, 'removeFromCart', data),
     purchase:       (data : PurchaseData) => track(config, 'purchase', data),
     setUser:        (data : UserData) => {
-        const oldUserId = config.user ? config.user.id : undefined;
-        config.user = config.user || { id: data.user.id };
-        config.user.id = data.user.id || config.user.id;
-        config.user.email = data.user.email || config.user.email;
-        config.user.name = data.user.name || config.user.name;
-        track(config, 'setUser', { oldUserId });
+        config.user = {
```
bluepnume: Mutating config feels weird here -- we're changing an object that we didn't create. Would it make sense to clone this object before we mutate it?

src/tracker-component.js
```diff
@@ -66,15 +66,40 @@ type Config = {|
...
+const getUserIdCookie = () : ?string => {
+    const userCookie = document.cookie.split(';').find(x => x.startsWith('paypal-cr-user'));
+    if (!userCookie) {
+        return;
+    }
+    return userCookie.split('=')[1];
+};
+
+const getUserId = () : ?string => {
+    return getUserIdCookie() || localStorage.getItem('user-id');
```
bluepnume: Same here.

src/tracker-component.js
bryanjenningz: do you mean changing it so it's like this instead?
```
config = {
    ...config,
    user: {
        ...config.user,
        email: data.user.email || ((config && config.user) || {}).email,
        name:  data.user.name || ((config && config.user) || {}).name
    }
};
```

src/tracker-component.js
bryanjenningz: I check for if the value returned is truthy here and if it's not truthy then I set localStorage so that it's always a value https://github.com/paypal/paypal-muse-components/pull/4/files#diff-bc1bab447af0f7377f922001d7fb757dR91
let me know if that's what you mean

src/tracker-component.js
bryanjenningz: here I'm setting the user id in localStorage, is that okay?

src/tracker-component.js
bluepnume: https://jonathanmh.com/use-ios-safari-localstorage-sessionstorage-private-mode/

src/tracker-component.js
bluepnume: 👍

src/tracker-component.js
bryanjenningz: thanks for the information, we change the implementation to only use cookies

src/cookie-utils.js
```diff
@@ -0,0 +1,23 @@
+/* @flow */
+
```
bryanjenningz: maybe I should just use js-cookie instead https://www.npmjs.com/package/js-cookie

PR #6, opened 2019-04-17
Add param to setUser to state if user is PP user

zacharyhill:
Will use later to check if email should be sent as branded/unbranded recovery email


zacharyhill:
Found a better way to do this

PR #11, opened 2019-05-14
Jetlore

zacharyhill:
Integrating Jetlore into our tracking SDK so a merchant could easily integrate with both Jetlore and our tracking SDK with only one integration.

There are some relevant comments surrounding this in the code, but sinon/rewire couldn't be imported due to the error `Uncaught Error: Module build failed: Error: Final loader (./node_modules/imports/index.js) didn't return a Buffer or String` -- which is why there is some repetitive manual mocking going on.

Also was getting an error surrounding the manual mocks `Cannot set property "prop" of #<Object> which has only a getter` which is why generateId has been exported as a default object and why there has been flow and eslint errors ignored surrounding that code.

Also sinon is added to the package.json even though it cannot be imported because karma throws this error if not: `ERROR [karma-server]: Server start failed on port 9876: Error: Cannot find module 'sinon/package.json'`


package.json
```diff
@@ -33,7 +33,8 @@
     "chai": "^4.2.0",
     "flow-bin": "^0.70.0",
     "grumbler-scripts": "^3",
-    "mocha": "^4"
+    "mocha": "^4",
+    "sinon": "^7.3.2"
```
songz: should sinon be removed?

zacharyhill: I just updated the description to include the error I'm getting when it's not removed. I get this error `ERROR [karma-server]: Server start failed on port 9876: Error: Cannot find module 'sinon/package.json'`

src/lib/jetlore.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
```
songz: could you add additional comment to state where this asset is pulled from? (Jetlore's asset url)

bluepnume: Would it make more sense to load this resource from paypalobjects rather than inlining it?

src/lib/jetlore.js
```diff
@@ -0,0 +1,18 @@
...
+let JL = {};
+!(function() { function o(o) { const t = this; t.access_token = o.cid, t.user_id = void 0 !== o.user_id && o.user_id != null && o.user_id !== '' && o.user_id.indexOf('/') < 0 ? o.user_id : 'undefined', t.feed_id = void 0 === o.feed_id || o.feed_id == '' ? 'any_feed' : o.feed_id, t.div = o.div, t.lang = o.lang, t.clicked = typeof o.clicked === 'function' ? o.clicked : function(o) {}, t.purchased = typeof o.purchased === 'function' ? o.purchased : function(o) {}, t.logToConsoleCookieName = 'Tracker_log_to_console', t.logToConsole = e._isConsoleCookieExists(t.logToConsoleCookieName), t.setCookies = void 0 !== o.setCookies && o.setCookies != null && o.setCookies, t.autoPageView = void 0 !== o.autoPageView && typeof o.autoPageView === 'boolean' && o.autoPageView, t.log(`Tracking initialized with parameters\n\t - access token:${ t.access_token  }\n\t - user id     :${  t.user_id  }\n\t - feed id     :${  t.feed_id  }\n\t - div         :${  t.div  }\n\t - lang        :${  t.lang }`), t.autoPageView && t.page_view(); }JL = { tracking(e) { JL.tracker = new o(e); } }, o.prototype.track = function(o) { const e = this,
```
bluepnume: This is going to de-opt the webpack build, since it has side effects at the module scope. Can we wrap this in a `getJetlore()` or something if we're going to inline it?

src/generate-id.js
```diff
@@ -1,4 +1,13 @@
...
-    [ Math.random(), Math.random() ].map(x => x.toString(16).slice(2)).join('');
+/*
+** the reason this is exported as a default object is to get rid of error
+** "Cannot set property "prop" of #<Object> which has only a getter"
+** - flow and eslint have to be disabled on whatever line this is used.
+*/
+
+// eslint-disable-next-line import/no-default-export
+export default {
+    generateId: () : string =>
```
bluepnume: If you like, you can `import { uniqueID } from 'belter/src';` -- this is the function we use throughout the rest of the sdk.

src/tracker-component.js
```diff
@@ -75,24 +92,67 @@ const getUserIdCookie = () : ?string => {
 
 const setRandomUserIdCookie = () : void => {
     const ONE_MONTH_IN_MILLISECONDS = 30 * 24 * 60 * 60 * 1000;
-    setCookie('paypal-user-id', generateId(), ONE_MONTH_IN_MILLISECONDS);
+    setCookie('paypal-user-id', generate.generateId(), ONE_MONTH_IN_MILLISECONDS);
+};
+
+
+export const getJetlorePayload = (type : string, payload : Object) : Object => {
```
bluepnume: If this needs to be exported, can we make it a child of `Tracker` rather than hanging a new function off of `paypal`?

src/lib/jetlore.js
songz: This is meant to be integrated with all the tracking events so Arvind Naidu prefers to bundle this in, unless you insist otherwise. Do you insist @bluepnume ?

src/lib/jetlore.js
```diff
@@ -0,0 +1,22 @@
...
+    return JL;
+}
+
+export default getJetlore();
```
bluepnume: This is still causing a module-level side-effect. There should be no direct function calls at the module level if webpack is going to tree-shake correctly.

PR #17, opened 2019-06-04
Add tests for dfp

zacharyhill:
Removed calls checking URL since we are already doing this when we decode encoded data to make sure it is the same as the data that is passed in from get-device-info


songz:
Merged this in because this is just tests.

PR #18, opened 2019-06-04
Change Jetlore JSONP function to use image tags instead of script tags

zacharyhill:
Doing this since we cannot get a CSP exception to hit Jetlore.

### DIFF
--- a/src/lib/jetlore.js
+++ b/src/lib/jetlore.js
@@ -311,12 +311,11 @@ const getJetlore = () => {
       },
     
       _jsonp: function(self, url) {
-          var scriptTag = document.createElement("script");
-          scriptTag.type = "text/javascript";
-          scriptTag.src=url;
-          scriptTag.async = true;
-          document.getElementsByTagName("head")[0].appendChild(scriptTag);
-          self.log("SENT JSONP request: " + url)
+        var imgTag = document.createElement("img");
+        imgTag.src=url;
+        imgTag.style.display = 'none'
+        document.getElementsByTagName("body")[0].appendChild(imgTag);
+        self.log("SENT JSONP request: " + url)
       },
     
       _getUTCTime: function(){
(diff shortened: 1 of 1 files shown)


src/lib/jetlore.js
```diff
@@ -311,12 +311,11 @@ const getJetlore = () => {
...
       _jsonp: function(self, url) {
-          var scriptTag = document.createElement("script");
-          scriptTag.type = "text/javascript";
-          scriptTag.src=url;
-          scriptTag.async = true;
-          document.getElementsByTagName("head")[0].appendChild(scriptTag);
-          self.log("SENT JSONP request: " + url)
+        var imgTag = document.createElement("img");
+        imgTag.src=url;
+        imgTag.style.display = 'none'
```
songz: in some browsers, display none might cause browser not to load it. Could you instead, style position fixed top left, with width and height of 1
https://stackoverflow.com/questions/12158540/does-displaynone-prevent-an-image-from-loading

src/lib/jetlore.js
```diff
@@ -311,12 +311,11 @@ const getJetlore = () => {
...
+        document.getElementsByTagName("body")[0].appendChild(imgTag);
```
songz: y not just document.body? `getElementsByTagName` will do a dom traversal

src/lib/jetlore.js
```diff
@@ -311,12 +311,11 @@ const getJetlore = () => {
...
+        self.log("SENT JSONP request: " + url)
```
songz: its not a jsonp request! `sent tracking request`

zacharyhill: That was the log they had in there originally.. Do you really want me to change it? The function is called _jsonp. This log goes to them not the browser.

songz: no

songz:
closing in preference for this: https://github.com/paypal/paypal-muse-components/pull/21

PR #25, opened 2019-06-19
Add identify method to tracker component

zacharyhill:
Setup identity route for getting token to send offer emails through CRNS route.

### DIFF
--- a/src/tracker-component.js
+++ b/src/tracker-component.js
@@ -273,12 +273,13 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
     };
     const identify = (cb? : function) => {
         let url;
+        const failurePayload = { success: false };
         if (config.paramsToTokenUrl) {
             url = config.paramsToTokenUrl();
         } else {
             url = 'https://paypal.com/muse/api/partner-token';
         }
-        const resp = window.fetch(url, {
+        return window.fetch(url, {
             method:      'POST',
             credentials: 'include',
             headers:     {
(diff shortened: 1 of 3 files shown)


src/tracker-component.js
```diff
@@ -265,10 +269,39 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
...
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+                merchantId: getMerchantID()[0],
+                clientId:   getClientID()
+            })
+        }).then(res => {
+            if (res.status !== 204) {
+                return cb({ success: false });
+            }
```
songz:
```
.then( res => {
  if (res.status !== 204) {
     cb({success.....})
     return null
  }
  return res.json()
}).then( data => {
  if (!data) return
  ....
})
```

src/tracker-component.js
```diff
@@ -265,10 +269,39 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
             trackers[type](data);
         }
     };
+    const identity = (cb : function) => {
```
songz: identify

src/tracker-component.js
```diff
@@ -267,10 +271,45 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
...
+            headers:     {
+                'Content-Type': 'application/json'
+            },
+            body: JSON.stringify({
+                merchantId: getMerchantID()[0],
+                clientId:   getClientID()
+            })
+        }).then(res => {
+            if (res.status !== 200) {
+                return cb ? cb(failurePayload) : false;
```
songz: `return false`

src/tracker-component.js
```diff
@@ -267,10 +271,45 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
...
+            }
+            return res.json();
+        }).then(data => {
+            if (!data) {
+                return failurePayload;
```
songz: `return cb ? ...`

PR #27, opened 2019-07-03
add getIdentity function for Tracker

insomnolent:


src/tracker-component.js
```diff
@@ -251,6 +276,19 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
         },
         setProperty: (data : PropertyData) => {
             config.properties = { ...config.properties, ...data };
+        },
+        getIdentity: (data: IdentityData) => {
+            // make url different for different envs later
+            const url = 'https://localhost.paypal.com:8443/muse/api/partner-token'
```
songz: this is getting merged into production, so this will not work. One thing you can do is take in an optional url in the `getIdentity` parameter.

insomnolent: what should the default url be if they don't pass in a url then?

src/tracker-component.js
```diff
@@ -107,6 +112,23 @@ const setCartCookie = (type, data) : void => {
     setCookie('paypal-cr-cart', JSON.stringify(data), sevenDays);
 };
 
+const getAccessToken = (url: string, mrid : string) : string => {
```
songz: url should have a default `/muse/....`

src/tracker-component.js
```diff
@@ -257,6 +279,13 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
         },
         setPropertyId: (id : string) => {
             config.propertyId = id;
+        },
+        getIdentity: (data: IdentityData, url?: string) => {
+            return getAccessToken(url, data.mrid)
+            .then(accessToken => data.onIdentification(accessToken))
```
songz: lets allow people to also use promises
```
if (data.onIdentification) {
  data.onIdentification( {getAccessToken: () => accessToken})
}
return accessToken
```

src/tracker-component.js
```diff
@@ -107,6 +114,22 @@ const setCartCookie = (type, data) : void => {
...
+        credentials: 'include',
+        headers:     {
+            'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({
+            mrid,
+            clientId: getClientID()
+        })
+    }).then(r => r.json()).then(data => {
+        return data.cr_token;
```
songz: Here's how CIQ was communicated to call this route: 
```
paypal.Tracker.getIdentity({
  mrid: 'oeuaoeuae',

   onIdentification: (data) => {

       console.log(data.getAccessToken()) // should print out the access_token

       /**

           Your code that saves accessToken.

       **/

   }
})
```

PR #31, opened 2019-07-16
Disable modal

zacharyhill:
Disable modal unless `beta` property is set to true on `cartRecovery` property in config object.

### DIFF
--- a/src/messaging-component.js
+++ b/src/messaging-component.js
@@ -51,8 +51,11 @@ const showExitModal = ({ cartRecovery }) => {
     return isRendered;
 };
 
-export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string } }>) => {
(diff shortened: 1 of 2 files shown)


src/messaging-component.js
```diff
@@ -51,8 +51,11 @@ const showExitModal = ({ cartRecovery }) => {
     return isRendered;
 };
 
-export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string } }>) => {
+export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string, beta? : boolean } }>) => {
     const config = args[0];
+    if (!config.cartRecovery || !config.cartRecovery.beta) {
```
zacharyhill: Also checked `config.cartRecovery` here since it seems pointless to setup the event listeners if they aren't going to be used to call anything.

PR #33, opened 2019-07-17
Warn users that Messaging component is deprecated in console

zacharyhill:


src/messaging-component.js
```diff
@@ -52,6 +52,7 @@ const showExitModal = ({ cartRecovery }) => {
 };
 
 export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string, beta? : boolean } }>) => {
+    console.warn('This component has been deprecated. You are probably looking for the "Messages" component'); // eslint-disable-line no-console
```
songz: `This component has been deprecated. Starting July 24, this functional will no longer be available and will result in an error. Please fix accordingly.`

PR #37, opened 2019-07-30
refactor(cart-storage): use localStorage to store cart and fix flow/lint errors

joshbeam:
## Problem

Currently we're using a first-party cookie to store JSON cart contents. We've encountered two problems with this:
1. One merchant's server was throwing errors when this cookie was sent to their server automatically
2. Cookies have size limits depending on the browser, so it's possible all the cart contents won't fit in the cookie.

## Solution

Use localStorage instead of a cookie. In order to maintain cookie expiration parity, I've also added a function during initialization to check an expiration value to determine if we should wipe the localStorage.

## Additional Notes

It looks like CI isn't enabled, and we don't have pre-commit hooks to run tests. When I ran tests locally, there were many flow/lint errors and two invalid test cases present. I've fixed those as well in this PR. We should activate CI and add pre-commit hooks.

### DIFF
--- a/src/messaging-component.js
+++ b/src/messaging-component.js
@@ -30,7 +30,8 @@ const showExitModal = ({ cartRecovery }) => {
         return false;
     }
     // don't show modal if user has no cart items
-    const cart = JSON.parse(window.localStorage.getItem('paypal-cr-cart') || '{}');
+    // Prefer localStorage first, fallback to cookie (backwards compatibility -- this check can be removed a couple weeks after deploy).
+    const cart = JSON.parse(window.localStorage.getItem('paypal-cr-cart') || getCookie('paypal-cr-cart') || '{}');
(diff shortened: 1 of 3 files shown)


src/messaging-component.js
```diff
@@ -30,7 +30,7 @@ const showExitModal = ({ cartRecovery }) => {
         return false;
     }
     // don't show modal if user has no cart items
-    const cart = JSON.parse(getCookie('paypal-cr-cart') || '{}');
+    const cart = JSON.parse(window.localStorage.getItem('paypal-cr-cart') || '{}');
```
songz: existing integrations should have their localStorage populated with existing cart from the cookie instead of `{}`.

songz:
approved, tested it on test merchant site.

PR #40, opened 2019-08-06
feat(auto-container-ids): Allow tracker init to query a property ID on the fly

zacharyhill:
Automatically get/set container ID for merchant.

Still working to add test cases for the following:
- setPropertyId is called instead of passing propertyId into config object (for backwards compatibility)
- check that correct propertyId goes into cart event
- check that track event queue is cleared when id resolves


src/tracker-component.js
```diff
@@ -254,6 +271,35 @@ const clearExpiredCart = () => {
...
+        if (paramsToPropertyIdUrl) {
+            url = paramsToPropertyIdUrl();
+        } else {
+            url = 'https://paypal.com/tagmanager/containers/xo';
+        }
+        fetch(`${ url }?mrid=${ merchantId }&url=${ encodeURIComponent(currentUrl) }`)
+            .then(res => {
+                if (res.status === 200) {
+                    return res;
+                }
```
joshbeam: Should we handle the case here if `res.status` isn't 200?

src/tracker-component.js
```diff
@@ -343,7 +389,18 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
             };
             track(config, 'setUser', { oldUserId: getUserIdCookie() });
         },
-        setPropertyId: (id : string) => {
+        setPropertyId: (id : string, automatic : boolean = false) => {
```
joshbeam: I'm not a fan of the pattern of having a single method that changes behavior based on a flag that's passed in, I think it increases complexity both on development and usage of the method. Would it make sense to just create another method? We can keep `setPropertyId` as it was originally, and then have a separate internal method for the new stuff.

src/tracker-component.js
```diff
@@ -254,6 +272,35 @@ const clearExpiredCart = () => {
...
+        window.fetch(`${ url }?mrid=${ merchantId }&url=${ encodeURIComponent(currentUrl) }`)
+            .then(res => {
+                if (res.status === 200) {
+                    return res;
+                }
+            })
+            .then(r => r.json()).then(container => {
+                window.localStorage.setItem(propertyIdKey, container.id);
+                resolve(container.id);
+            });
```
songz: add a catch

src/tracker-component.js
songz: could we TD this to add proper logging?

src/tracker-component.js
zacharyhill: Done. Not sure why github isn't marking this as outdated.

src/tracker-component.js
```diff
@@ -246,6 +256,14 @@ const track = <T>(config : Config, trackingType : TrackingType, trackingData : T
     }
 };
 
+const clearTrackQueue = config => {
```
joshbeam: Does this work? Maybe I have the wrong understanding. I tried to replicate it in the console:

```
let foo = [1,2,3,4,5];
const doIt = () => {
  foo.forEach(int => {
     console.log(int) // equivalent to track(whatever)

     foo.shift();
  })

   console.log(foo)
}

doIt()
VM257:4 1
VM257:4 3
VM257:4 5
VM257:9 (2) [4, 5] // the `foo` queue still has some items left in it
```

It looks like it only fires some events, but that some would be left in the tracking queue?

src/tracker-component.js
```diff
@@ -246,6 +255,14 @@ const track = <T>(config : Config, trackingType : TrackingType, trackingData : T
...
+    // eslint-disable-next-line array-callback-return
+    window.trackEventQueue = window.trackEventQueue.filter(eventData => {
```
joshbeam: Thanks for the fix 👍 Can we add a test for this if we don't have one? I think the behavior would've been caught with that

src/tracker-component.js
zacharyhill: Good catch. Pushed a fix. Unfortunately that's a test case I couldn't get to work due to not being able to mock anything (which is why it was added to the window). I'm not sure where/when this got messed up, but I added an issue for this repo to hopefully take care of this issue later so we do not have this issue in the future.

PR #41, opened 2019-08-06
fix(remove-from-cart): removeFromCart modifies storage and allows multiple removals

M-OC:
- Individual items now removed from localstorage. 
Format: 
`{id:555, quantity:4}` <- removes up to 4 instances of items with id '555' 
`{id:555}` <- removes all instances of items with id '555'

- composeCart accepts 'remove' call. Uses switch.

### DIFF
--- a/src/lib/compose-cart.js
+++ b/src/lib/compose-cart.js
@@ -5,12 +5,10 @@ export const removeFromCart = (items, currentItems = []) => {
         if (item.quantity === undefined) {
             return accumulator.filter(curItem => curItem.id !== item.id);
         }
+
+        let quantity = parseInt(item.quantity, 10);
   
-        if (typeof item.quantity !== 'number') {
-            throw new TypeError('quantity must be a number');
-        }
-  
-        while (item.quantity > 0) {
+        while (quantity > 0) {
             const index = accumulator.findIndex(curItem => curItem.id === item.id);
   
             if (index === -1) {
@@ -18,7 +16,7 @@ export const removeFromCart = (items, currentItems = []) => {
             }
   
             accumulator.splice(index, 1);
-            item.quantity -= 1;
+            quantity -= 1;
         }
   
         return accumulator;
--- a/src/lib/compose-cart.js
+++ b/src/lib/compose-cart.js
@@ -1,0 +1,30 @@
+/* @flow */
+// $FlowFixMe
+export const removeFromCart = (items, currentItems) => {
+    return items.reduce((accumulator, item) => {
+        if (item.quantity === undefined) {
+            return accumulator.filter(curItem => curItem.id !== item.id);
+        }
+  
+        if (typeof item.quantity !== 'number') {
+            throw new TypeError('quantity must be a number');
+        }
+  
+        while (item.quantity > 0) {
+            const index = accumulator.findIndex(curItem => curItem.id === item.id);
+  
+            if (index === -1) {
+                break;
+            }
+  
+            accumulator.splice(index, 1);
+            item.quantity -= 1;
+        }
+  
+        return accumulator;
+    }, currentItems);
+};
+// $FlowFixMe
+export const addToCart = (items, currentItems = []) => {
+    return [ ...currentItems, ...items ];
+};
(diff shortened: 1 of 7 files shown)


src/tracker-component.js
```diff
@@ -330,7 +335,11 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
 
             return trackCartEvent(config, 'setCart', newCart);
         },
-        removeFromCart: (data : RemoveCartData) => trackCartEvent(config, 'removeFromCart', data),
+        removeFromCart: (data : RemoveCartData) => {
+            const newCart = composeCart('remove', data);
```
joshbeam: Do we need test coverage for this portion? Making sure that removeFromCart modifies the localStorage as expected

M-OC: I did some manual testing, it works as-expected for subsequent calls.

You're right though. I'll write some tests for the 'parent' composeCart method and throw a bunch of different get/set/remove calls at it to make sure everything is squared away.

src/tracker-component.js
```diff
@@ -330,7 +336,11 @@ export const Tracker = (config? : Config = defaultTrackerConfig) => {
 
             return trackCartEvent(config, 'setCart', newCart);
         },
-        removeFromCart: (data : RemoveCartData) => trackCartEvent(config, 'removeFromCart', data),
+        removeFromCart: (data : RemoveCartData) => {
+            composeCart('remove', data);
+
+            trackCartEvent(config, 'removeFromCart', data);
```
M-OC: Please give this line a look & feedback.

You'll notice that 'data' here is different from the calls to 'add' & 'set':
- 'add' & 'set' return the _entire_ cart in it's new state. This means if the cart had one item in it, and another was added in a call to 'add', the cart event will be called with two items.
- 'remove' works differently. The data will reflect only the items that were removed.

joshbeam: What data did `removeFromCart` return before you made this change? I think as long as we maintain parity, then we should be fine

M-OC: It's the same as it was. Data is passed from the event directly to the tracking event.

src/lib/compose-cart.js
```diff
@@ -0,0 +1,30 @@
+/* @flow */
+// $FlowFixMe
+export const removeFromCart = (items, currentItems = []) => {
+    return items.reduce((accumulator, item) => {
+        if (item.quantity === undefined) {
+            return accumulator.filter(curItem => curItem.id !== item.id);
+        }
+  
+        if (typeof item.quantity !== 'number') {
```
shivinkapur: are we certain quantity is expected to be a `number`?

shivinkapur: https://github.com/paypal/paypal-muse-components/blob/e3c8056fe741598f4cc82878f7a49efe82e7a7c8/src/tracker-component.js#L25

src/tracker-component.js
```diff
@@ -107,28 +108,33 @@ const setRandomUserIdCookie = () : void => {
...
+        break;
+    case 'set':
+        _data.items = data.items;
+        break;
+    case 'remove':
+        _data = { ...cart, ...data };
+        _data.items = removeFromCart(data.items, currentItems);
+        break;
+    default:
+        throw new Error('invalid cart action');
```
shivinkapur: _*nitpick*_ its only our code calling reaching this code. do we still want to throw? we can have default just `return` instead

songz: throw might break merchant site.

src/lib/compose-cart.js
songz: for this..... what if we just made a best effort decision?
```
const quantity = parseInt( item.quantity, 10 ) || 0
```

src/tracker-component.js
M-OC: Calling this method with a string other than those three would be a programmer error that would break things anyway. I don't see the issue with being explicit about what went wrong.

src/lib/compose-cart.js
M-OC: What would the advantage be of accepting 'quantity' as a string?

songz: not much.... `removeFromCart( [{title: 'shoes', quantity: '2'}] )` feels strange as a developer. So for consistency, we should either change this `if` statement or change the `type` to number

PR #42, opened 2019-08-06
Remove iOS disable

songz:
This PR will allow safari tracking events to go through. For reference:
https://github.com/paypal/paypal-muse-components/commit/b3e76554fadd72ad24b6a900b99b8ff75af08815


songz:
@shivinkapur it does not affect it, because view is an empty function: https://github.com/paypal/paypal-muse-components/pull/42/files#diff-bc1bab447af0f7377f922001d7fb757dR313

PR #52, opened 2019-08-29
Set user

M-OC:
- extracts types, constants into separate files
- extracts 'track', 'getPropertyId' and some minor methods into separate files
- check for 'propertyId' no longer occurs within 'track' method
- 'setUser' no longer accepts 'undefined'
- 'setUser' now accepts `{user: {email: 'foo'}}` and `{email: 'foo'}`


.eslintrc.js
```diff
@@ -420,7 +420,7 @@ module.exports = {
         'import/exports-last': 'off',
         'import/group-exports': 'off',
         'import/no-cycle': 'error',
-        'import/no-default-export': 'error',
+        'import/no-default-export': 'off',
```
M-OC: We were already ignoring this rule elsewhere.

src/declarations.js
```diff
@@ -1,5 +1,5 @@
 /* @flow */
 
-import type { MuseGlobalType } from './types';
+import type { MuseGlobalType } from './lib/types';
```
songz: would it make sense to create a `types` folder and put `MuseGlobalType` in `index.js` for now until we have more types? When I think about `lib` usually I think helper libraries and things like that. I think `types` should be in a folder of its own.

src/lib/track.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+    const user = {
+        ...config.user,
+        id: getUserIdCookie()
+    };
+
+    // remove null, undefined values
+    user.id || delete user.id
+    user.email || delete user.email
+    user.name || delete user.name
```
M-OC: The above is new.

src/lib/get-property-id.js
```diff
@@ -0,0 +1,33 @@
+import { getClientID, getMerchantID } from '@paypal/sdk-client/src';
+
+export const getPropertyId = ({ paramsToPropertyIdUrl }) => {
```
M-OC: This file is unchanged.

src/lib/compose-cart.js
```diff
@@ -26,3 +33,39 @@ export const removeFromCart = (items, currentItems = []) => {
 export const addToCart = (items, currentItems = []) => {
     return [ ...currentItems, ...items ];
 };
+// $FlowFixMe
+export const composeCart = (type, data) => {
```
M-OC: This method now uses 'constants' imported above.  Otherwise unchanged.

src/lib/cookie-utils.js
```diff
@@ -21,3 +22,12 @@ export const setCookie = (cookieName : string, cookieValue : string, expirationM
     const expires = `expires=${ d.toUTCString() }`;
     document.cookie = `${ cookieName }=${ cookieValue }; Path=/; ${ expires }`;
 };
+
+export const getUserIdCookie = () : ?string => {
```
M-OC: These methods were removed from 'tracker-component'. They have not been altered. (note: ONE_MONTH_IN_MILLISECONDS should probably be in constants)

src/lib/constants.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+export default {
+    'sevenDays': 6.048e+8,
```
songz: Personally, (nitpick) I think its easier to understand if you wrote it like this: `sevenDays: 1000*60*60*24*7`

😆

src/lib/track.js
```diff
@@ -0,0 +1,52 @@
...
+    // to whatever you want it to be based on the trackingType string and data object.
+    // This can be useful for testing purposes, this feature won't be used by merchants.
+    if (config.paramsToBeaconUrl) {
+        img.src = config.paramsToBeaconUrl({ trackingType, data });
+    } else {
+        img.src = `https://www.paypal.com/targeting/track/${ trackingType }?data=${ encodeData(data) }`;
+    }
+
+    if (document.body) {
+        document.body.appendChild(img);
```
songz: Just noticed that we are not removing the image element. Could you add a TODO here to remove the image after it has loaded?

src/lib/track.js
```diff
@@ -0,0 +1,52 @@
...
+    if (!getUserIdCookie()) {
+        setRandomUserIdCookie();
+    }
+    const user = {
+        ...config.user,
+        id: getUserIdCookie()
+    };
+
+    // remove null, undefined values
+    user.id || delete user.id
```
songz: this is impossible right? Because right above you are setting userIdCookie

src/lib/track.js
songz: I thought setting `email` and `name` field to `null` or `''` will reset it on our back end? We want these values to get passed down the pipeline to get processed as is, so maybe we should not delete these fields?

src/tracker-component.js
```diff
@@ -4,150 +4,31 @@ import 'whatwg-fetch'; // eslint-disable-line import/no-unassigned-import
...
+    IdentityData,
+    JetloreConfig,
+    Config
+} from './lib/types';
+
+const {
+    accessTokenUrl,
+    storage,
+    defaultTrackerConfig
+} = constants;
```
songz: why not just `import { accessokenUrl .... } from './lib/constants'`

src/tracker-component.js
```diff
@@ -214,59 +95,26 @@ const getJetlorePayload = (type : string, options : Object) : Object => {
 
 let trackEventQueue = [];
 
-const track = <T>(config : Config, trackingType : TrackingType, trackingData : T) => {
+export const clearTrackQueue = (config : Config) => {
+    // $FlowFixMe
+    return trackEventQueue.filter(([ trackingType, trackingData ]) => { // eslint-disable-line array-callback-return
```
songz: nitpick: much cleaner to use a forEach and then `return []` imo

src/tracker-component.js
```diff
@@ -214,59 +95,26 @@ const getJetlorePayload = (type : string, options : Object) : Object => {
...
+        track(config, trackingType, trackingData);
```
songz: question, does `track` also fire off JL events?

src/tracker-component.js
joshbeam: Agreed. This was Zach's original implementation, and looks like it's just moved in the file. Maybe we can add a TODO instead to change this method?

songz: 👍

M-OC: added

src/tracker-component.js
M-OC: It uses a different 'tracker' method. See "trackEventByType" & "getJetlore".

src/lib/track.js
M-OC: Done.

src/declarations.js
M-OC: Agreed, but I'd prefer to do that + additional refactoring in another PR if possible.

src/tracker-component.js
M-OC: Each key would have to be exported separately in that case.

src/tracker-component.js
songz: https://github.com/paypal/paypal-muse-components/issues/53

PR #58, opened 2019-08-30
Create 'setCartId' method

M-OC:


src/lib/track.js
```diff
@@ -10,6 +11,16 @@ import type {
 
 export const track = <T>(config : Config, trackingType : TrackingType, trackingData : T) => {
     const encodeData = data => encodeURIComponent(btoa(JSON.stringify(data)));
+    let cartId;
+
+    try {
+        // $FlowFixMe
+        cartId = getCartId().cartId;
+    } catch (err) {
```
joshbeam: [nit] I assume the catch is here because getCartId() could return `null`, and `(null).cartId` would throw an error? Seems a little unintuitive, I might do `cartId = getCartId() ? getCartId().cartId : createNewCartId().cartId`

M-OC: Agreed. It isn't intuitive.

With the current implementation this _should_ never happen. Ideally 'track' should just read from the cartId in localstorage. If 'track' is getting called and it doesn't exist it's basically a programmer error. I wanted to surface that in some way.

M-OC: Done. Created a new method 'getOrCreateCartId' that explicitly get or sets.

src/lib/local-storage-utils.js
```diff
@@ -0,0 +1,53 @@
+/* @flow */
+import constants from './constants';
+import generate from './generate-id';
+
+const { storage, sevenDays } = constants;
+
+/* Returns an existing cartId or null */
+export const getCartId = () => {
+    const storedValue = window.localStorage.getItem(storage.paypalCrCart);
```
songz: this function should return `valid existing cartId`, not just cartId right? I don't think it should be developer logic to handle validity of cartIds.

src/lib/local-storage-utils.js
```diff
@@ -0,0 +1,53 @@
...
+};
+
+/* Returns an existing cartId or creates a new one */
+export const getOrCreateCartId = () => {
+    return getCartId() || createNewCartId();
+};
+
+/* Generates a new cartId if the existing value is older than 7 days.
+Returns the current, up-to-date cartId in either case */
+export const resetExpiredCartId = () => {
```
songz: I think `getOrCreateValidCartId` is a better name for this, because you are not only reseting expired cart, but also returning a valid cartId.

songz: if what I said above makes sense (re. naming `getOrCreateValidCartId`), then the `getOrCreateCartId` function has little to no value.

src/lib/local-storage-utils.js
```diff
@@ -0,0 +1,53 @@
+/* @flow */
+import constants from './constants';
+import generate from './generate-id';
+
+const { storage, sevenDays } = constants;
+
+/* Returns an existing cartId or null */
+export const getCartId = () => {
```
songz: this function is exported for the sole reason for testing? Its not used anywhere, and I can't think of a use case other than being a helper function in here. What do you think about not exporting it altogether (and then remove the test for `getCartId`

src/lib/track.js
```diff
@@ -10,6 +11,7 @@ import type {
 
 export const track = <T>(config : Config, trackingType : TrackingType, trackingData : T) => {
     const encodeData = data => encodeURIComponent(btoa(JSON.stringify(data)));
+    const cartId = getOrCreateCartId().cartId;
```
songz: I think you want a nonExpired cartId here, so perhaps the `getOrCreateValidCartId`, or `resetExpiredCartId` is more relevant here?

src/tracker-component.js
```diff
@@ -168,7 +156,7 @@ export const Tracker = (config? : Config = {}) => {
         console.log('PayPal Shopping: debug mode on.');
     }
     
-    clearExpiredCart();
+    resetExpiredCartId();
```
songz: If `getOrCreateValidCartId` takes care of expiring cartIds, then this is no longer necessary.

src/lib/local-storage-utils.js
M-OC: This is used by both 'resetExpiredCartId' and 'getOrCreate'. It's exported for the sake of testing.

src/lib/local-storage-utils.js
M-OC: This is functionally different from 'getOrCreate' in that it checks the 'createdAt' value associated with the cartId.

The only reason 'getOrCreate' exists is to handle a potential programmer error. Ideally track should have nothing to do with initializing/setting up a cart id.

src/lib/track.js
M-OC: Honestly, I'd rather this didn't exist at all. The 'track' method shouldn't concern itself with this step in the first place. By allowing this to fail gracefully I think we're opening the door to people inadvertently botching the setup process without realizing it.

src/tracker-component.js
M-OC: Done.

PR #60, opened 2019-09-03
Input validation

M-OC:


src/tracker-component.js
```diff
@@ -218,6 +236,14 @@ export const Tracker = (config? : Config = {}) => {
             return event;
         },
         setUser: (data : UserData) => {
+            try {
+                validateUser(data);
```
songz: should be `data.user || data`

songz: Screen Shot 2019-09-03 at 12 41 36 PM

left: v2. 
Right: v1

M-OC: Screen Shot 2019-09-03 at 1 14 48 PM

PR #61, opened 2019-09-04
Cart total

M-OC:
- 'total' now required by add/set/remove cart methods.
- creates 'normalizer' functions for various sdk methods. These are meant to handle differences between versions of the sdk.
- logs a warning when a 'normalizer' recognizes a deprecated implementation.
- updates tests.


src/lib/deprecated-input-normalizers.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+/* sets cartTotal to total. Warns in the event a 'deprecated' value is passed */
+export const addToCartNormalizer = (input : any) => {
+    if (input) {
+        if (input.total !== undefined) {
+            // eslint-disable-next-line no-console
+            console.warn('"total" has been deprecated. use "cartTotal" instead');
+        } else if (input.cartTotal) {
+            input.total = input.cartTotal;
```
joshbeam: Why not `input.cartTotal = input.total`? I would assume we should be sending the new keys to the back-end, not the old ones (basically looks like you're normalizing v2 into v1 here, instead of v1 into v2).

joshbeam: Had an offline discussion -- we're using "total" on the back-end, and changing that would require a change to multiple services and the database. Max said he'll put a code comment here explaining that.

PR #64, opened 2019-09-05
user id

M-OC:
- userId moved from cookies to localstorage (expire after one month)
- setUser overwrites user id when one is passed
- null removes existing user id with a randomized one we generate

### DIFF
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -1,7 +1,7 @@
 /* @flow */
 export default {
-    'sevenDays': 6.048e+8,
-    'oneMonth': 2592000000,
+    'sevenDays': 1000 * 60 * 60 * 24 * 7,
+    'oneMonth': 1000 * 60 * 60 * 24 * 30,
     'accessTokenUrl': 'https://www.paypal.com/muse/api/partner-token',
     'storage': {
         'paypalCrUser': 'paypal-cr-user',
--- a/src/lib/constants.js
+++ b/src/lib/constants.js
@@ -1,8 +1,10 @@
 /* @flow */
 export default {
     'sevenDays': 6.048e+8,
+    'oneMonth': 2592000000,
     'accessTokenUrl': 'https://www.paypal.com/muse/api/partner-token',
     'storage': {
+        'paypalCrUser': 'paypal-cr-user',
         'paypalCrCart': 'paypal-cr-cart'
     },
     'defaultTrackerConfig': {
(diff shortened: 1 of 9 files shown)


src/tracker-component.js
```diff
@@ -178,6 +181,12 @@ export const Tracker = (config? : Config = {}) => {
     
     getOrCreateValidCartId();
 
+    if (config && config.user && config.user.id) {
+        setUserId(config.user.id);
+    } else {
+        getOrCreateValidUserId();
```
joshbeam: let's wrap in a try/catch since it uses JSON.parse under the hood

src/lib/constants.js
```diff
@@ -1,8 +1,10 @@
 /* @flow */
 export default {
     'sevenDays': 6.048e+8,
+    'oneMonth': 2592000000,
```
songz: nit: might be better if you just did: `1000*60*60*24*30`

songz: instead of calculating it and putting the calculated value here :)

src/lib/local-storage-utils.js
```diff
@@ -45,3 +45,45 @@ export const getOrCreateValidCartId = () => {
...
+
+    if (storedValue) {
+        return JSON.parse(storedValue);
+    }
+
+    return null;
+};
+
+/* Returns an existing, valid userId or creates a new one */
+export const getOrCreateValidUserId = () => {
```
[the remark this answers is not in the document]
songz: invalid = userId > 30 days.

src/lib/constants.js
M-OC: Ah, yeah maybe we should just get away from the gigantic, cryptic numbers.

src/lib/local-storage-utils.js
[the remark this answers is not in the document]
M-OC: "invalid" would be expired. The extra word is there to indicate that it isn't necessarily a pure get-or-create, since a value might already exist but we consider it expired.

src/tracker-component.js
M-OC: Fails gracefully now.

src/lib/constants.js
joshbeam: can we represent it as a roman numeral MMMCMXCIXCMXCIX

M-OC: I'll drop what I'm doing and get started.

PR #67, opened 2019-09-23
cancel cart no longer requires arguments

M-OC:
- prevents a bug encountered in the paypal sdk 'cancelCart' method. (No longer requires arguments, was looking up properties of undefined.)


test/tracker-component.test.js
```diff
@@ -386,11 +386,11 @@ describe('paypal.Tracker', () => {
         const tracker = Tracker({ currencyCode: 'COWRIESHELLS', user: { email, name: userName } });
         tracker.setPropertyId(propertyId);
         expect(appendChildCalls).toBe(0);
-        tracker.cancelCart({ cartId: '__test__cartId' });
+        tracker.cancelCart();
```
songz: if we left the test as is, does anything break? (concerned about existing v1 integrations).

PR #70, opened 2019-09-26
fix duplicate image element bug

M-OC:
Addresses a bug within the 'track' function.

Currently every time the 'track' method is called a completely new image element is created and attached to the document body. Over time a large number of elements can build up.

This PR modifies the 'track' method so that new image elements replace old ones.


joshbeam:
What is this bug? Can we create a ticket to track it?

src/lib/track.js
```diff
@@ -3,21 +3,36 @@ import { getClientID, getMerchantID } from '@paypal/sdk-client/src';
...
 
+const { imgElementId } = constants;
+
+const updateImgSrc = src => {
+    const oldImg = document.getElementById(imgElementId);
+
+    const newImg = document.createElement('img');
+    newImg.id = imgElementId;
+    newImg.style.display = 'none';
+    newImg.src = src;
```
songz: instead of tracking an element `id` (another thing to keep track of) and traversing the dom, why not just create and remove?
```
const newImg = document.createElement('img')
newImg.src....
newImg.onload = function() {
  newImg.remove()
}
newImg.onerror = function() {
  // possibly log fpti event
  newImg.remove()
}
```

PR #71, opened 2019-10-02
Trigger 'pageView' fpti event.

M-OC:
- pageView event fires fpti event
- pageView events will be pushed onto the 'trackEventQueue' and fired at the time the site 'propertyId' loads.
-  cartId, fpti keys related to identity not yet implemented.

### DIFF
--- a/src/lib/fpti.js
+++ b/src/lib/fpti.js
@@ -22,13 +22,12 @@ const sendBeacon = (src, data) => {
     beaconImage.src = `${ src }?${ query }`;
 };
 
+// removes empty strings, `undefined`, `null`, and `NaN` from fpti event
 const filterFalsyValues = source => {
     const result = {};
 
     for (const key in source) {
         if (source.hasOwnProperty(key)) {
-            // Only add to tracking source is values are truthy, `false`, or `0`
-            // (reject empty string, `undefined`, `null`, and `NaN`)
             if (source[key] || source[key] === false || source[key] === 0) {
                 result[key] = source[key];
             }
@@ -43,7 +42,10 @@ const resolveTrackingData = (config, data) => {
 
     return {
         e: 'im',
+        comp: 'ppshoppingsdk',
         page: `ppshopping:${ data.eventName }`,
+        t: new Date().getTime(),
+        g: new Date().getTimezoneOffset(),
         ...deviceInfo,
         ...config,
         ...data
@@ -121,21 +123,18 @@ const resolveTrackingVariables = (data) => ({
     pgrp: data.page,
 
     // Legacy impression event
-    e: data.e
+    e: data.e,
 
+    // Timestamp
+    t: data.t,
+
+    // Timestamp relative to user
+    g: data.g
 });
 
 export default (config, data = {}) => {
     const fptiServer = 'https://t.paypal.com/ts';
+    const trackingVariables = resolveTrackingVariables(resolveTrackingData(config, data));
 
-    const resolvedData = resolveTrackingVariables(resolveTrackingData(config, data));
-
-    const trackVariables = {
-        ...resolvedData,
-        e: 'im',
-        t: new Date().getTime(),
-        g: new Date().getTimezoneOffset()
-    };
-
-    sendBeacon(fptiServer, filterFalsyValues(trackVariables));
+    sendBeacon(fptiServer, filterFalsyValues(trackingVariables));
 };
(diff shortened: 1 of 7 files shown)


src/lib/fpti.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const beaconImage = new window.Image();
+    beaconImage.src = `${ src }?${ query }`;
+};
+
+// removes empty strings, `undefined`, `null`, and `NaN` from fpti event
+const filterFalsyValues = source => {
+    const result = {};
+
+    for (const key in source) {
```
songz: https://stackoverflow.com/questions/30326452/why-is-object-keys-faster-than-hasownproperty

Why not use `Object.keys(source)` instead of a `for` and `hasOwnProperty`? Applies to all functions.

src/lib/fpti.js
```diff
@@ -0,0 +1,140 @@
...
+    for (const key in data) {
+        if (data.hasOwnProperty(key)) {
+            query.push(`${ encodeURIComponent(key) }=${ encodeURIComponent(data[key]) }`);
+        }
+    }
+
+    query = query.join('&');
+
+    const beaconImage = new window.Image();
+    beaconImage.src = `${ src }?${ query }`;
```
songz: add event listener to remove these beacon images as well

src/lib/get-device-info.js
```diff
@@ -17,6 +17,22 @@ const getBrowserHeight = () => window.innerHeight;
 
 const getBrowserWidth = () => window.innerWidth;
 
+const getScreenHeight = () => {
+    try {
+        return window.screen.height;
+    } catch (err) {
+        return null;
+    }
```
songz: could we import `lodash` and then just do `_.get(window, 'screen.height', null)` (might be wrong, I forgot the exact API)

M-OC: Client doesn't currently include `lodash` unfortunately. We only have one dependency atm. I think the goal is to keep things as slim as possible.

src/lib/fpti.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    query = query.join('&');
+
+    const beaconImage = new window.Image();
+    beaconImage.src = `${ src }?${ query }`;
+};
+
+// removes empty strings, `undefined`, `null`, and `NaN` from fpti event
+const filterFalsyValues = source => {
```
joshbeam: Trying to think.. maybe we'd actually want to know on the serverside if a value like undefined or NaN was passed. Could help w/ debugging, etc. What do you think?

src/lib/fpti.js
```diff
@@ -0,0 +1,134 @@
...
+    sinfo: data.eventData,
+
+    // Legacy value for filtering events in Herald
+    page: data.page,
+
+    // Legacy value for filtering events in Herald
+    pgrp: data.page,
+
+    // Legacy impression event
+    e: data.e,
```
joshbeam: Not necessary to change right now, but for other events besides pageView, we don't necessarily want to always include `e=im`. Maybe make a comment in the code so we don't forget about it when we implement the rest of the client-side fpti events?

src/lib/fpti.js
```diff
@@ -0,0 +1,134 @@
...
+    bn_code: getPartnerAttributionID(),
+
+    // Event Name
+    event_name: data.eventName,
+
+    // Event Type
+    event_type: data.eventType,
+
+    // Event Data
+    sinfo: data.eventData,
```
joshbeam: Just double-checking: sinfo is expected to be a stringified object. Will this be the case here? (I believe if we were to just send a plain object it would be encoded server side like `[object Object]`, which we don't want)

src/lib/fpti.js
```diff
@@ -0,0 +1,134 @@
...
+    ru: data.location,
+
+    // Identification confidence score
+    confidence_score: data.confidenceScore,
+
+    // Identification type returned by VPNS
+    identifier_used: data.identifierUsed,
+
+    // Unverified encrypted customer account number
+    unverified_cust_id: data.userEAN,
```
joshbeam: These vpns values won't be present until we integrate identity on the client, right?

M-OC: Yes.

src/lib/fpti.js
M-OC: Good catch.

src/lib/fpti.js
M-OC: Ah, I'd rather explicitly log error conditions when we find them rather than allowing them to clutter here. Also we could technically identify a problem just by the fact certain values would be missing.

M-OC:
Screen Shot 2019-10-04 at 12 48 32 PM

PR #72, opened 2019-10-04
feat(store-user-ids): store merchant provided user ID and generated ID

joshbeam:
Before this change, we were storing `paypal-cr-user = { userId: asdlkjad_sflkjsdf }` in localStorage, and the userId key would be either the user ID we generated randomly, or the merchant provided user ID.

In this PR, we'll keep that behavior, but also add the keys `generatedUserId` and `merchantProvidedUserId` to that storage so that we keep track of all three values. Note that the behavior of userId being overloaded is not changed so that we maintain compatibility with existing methods/functions w/o having to do a big refactor.

Basic logic:

1. If there is a legacy user storage, at the time of load, the SDK will set `generatedUserId = userId` and keep the existing expiry
2. If we generate a new user ID, it will set both `userId` and `generatedUserId`
3. If we call `setUser`, it will set both `userId` and `merchantProvidedUserId`
4. For both 2 and 3, a new shared `createdAt` will be set to `Date.now`

Next steps:

This PR is only concerned with maintaining these values in local storage. The next PR will be concerned with sending all of these IDs to our tracking proxy.

### DIFF
--- a/src/lib/local-storage-utils.js
+++ b/src/lib/local-storage-utils.js
@@ -46,23 +46,54 @@ export const getOrCreateValidCartId = () => {
     return storedValue;
 };
 
-/* Sets a new userId to expire in one month */
-export const setUserId = (userId : string) => {
-    const storedValue = {
-        userId,
-        createdAt: Date.now()
-    };
+export const getUserStorage = () => {
+    let userStorage = window.localStorage.getItem(storage.paypalCrUser) || '{}';
 
-    window.localStorage.setItem(storage.paypalCrUser, JSON.stringify(storedValue));
+    try {
+        userStorage = JSON.parse(userStorage);
+    } catch (err) {
+        userStorage = {};
+    }
 
-    return storedValue;
+    return userStorage;
 };
 
-/* Generates a random userId to expire in one month */
-export const createNewUserId = () => {
-    const userId = `${ generate.generateId() }`;
+export const setUserStorage = (userStorage : Object, expiry? : Date) => {
+    userStorage.createdAt = expiry || Date.now();
 
-    return setUserId(userId);
+    window.localStorage.setItem(storage.paypalCrUser, JSON.stringify(userStorage));
+};
+
+// Generates a random user ID.
+// This method will set the userId field and generatedUserId field.
+export const setGeneratedUserId = (id? : string, expiry? : Date) => {
+    const userStorage = getUserStorage();
+
+    const userId = id || `${ generate.generateId() }`;
+
+    userStorage.userId = userId;
+
+    userStorage.generatedUserId = userId;
+
+    setUserStorage(userStorage, expiry);
+
+    return userStorage;
+};
+
+// Set the merchant provided user ID to the userId field and
+// the merchantProvidedUserId field.
+export const setMerchantProvidedUserId = (id : string) => {
+    const userStorage = getUserStorage();
+
+    userStorage.merchantProvidedUserId = id;
+
+    // The `userId` key will be merchantProvidedUserId || sdkGeneratedUserId for compatibility
+    // with the existing methods that send events.
+    userStorage.userId = id;
+
+    setUserStorage(userStorage);
+
+    return userStorage;
 };
 
 /* Returns a userId if one exists */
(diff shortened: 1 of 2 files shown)


src/tracker-component.js
```diff
@@ -302,10 +302,12 @@ export const Tracker = (config? : Config = {}) => {
...
-            if (data.id) {
-                setUserId(data.id);
-            } else if (data.id === null) {
-                createNewUserId();
+            if (data.id || data.id === null) {
+                setMerchantProvidedUserId(data.id);
+
+                if (data.id === null) {
+                    setGeneratedUserId();
+                }
```
songz: could this logic be simpler here? 
```
if (data.id || data.id === 0) { // because 0 can be a valid id.
  setMerchantProvidedUserId(data.id)
} else { // I think else is better, it takes care of `undefined` cases. I don't think merchants will pass in null.
   setGeneratedUserId()
}
```

src/lib/local-storage-utils.js
```diff
@@ -76,13 +107,22 @@ export const getUserId = () => {
...
 
-/* Returns an existing, valid userId or creates a new one */
+/* Returns an existing, valid userId or creates a new one if it doesn't exist or is expired */
 export const getOrCreateValidUserId = () => {
     const storedValue = getUserId();
     const now = Date.now();
 
     if (!storedValue || ((now - storedValue.createdAt) > oneMonth)) {
-        return createNewUserId();
+        // REVIEW: should we also clear the merchantProvidedUserId?
```
songz: per our conversation, let's just never clear userIds.

src/tracker-component.js
```diff
@@ -186,16 +186,16 @@ export const Tracker = (config? : Config = {}) => {
...
-        } else {
-            getOrCreateValidUserId();
+            setMerchantProvidedUserId(config.user.id);
         }
     } catch (err) {
         // eslint-disable-next-line no-console
         console.error(err.message);
         createNewCartId();
-        createNewUserId();
+        setGeneratedUserId();
```
songz: do we need both ` setGeneratedUserId()` and `getOrCreateValidUserId`?

src/lib/local-storage-utils.js
```diff
@@ -46,23 +46,54 @@ export const getOrCreateValidCartId = () => {
...
+// the merchantProvidedUserId field.
+export const setMerchantProvidedUserId = (id : string) => {
+    const userStorage = getUserStorage();
 
-    return setUserId(userId);
+    userStorage.merchantProvidedUserId = id;
+
+    // The `userId` key will be merchantProvidedUserId || sdkGeneratedUserId for compatibility
+    // with the existing methods that send events.
+    userStorage.userId = id;
```
songz: nah we don't need this.

src/lib/local-storage-utils.js
```diff
@@ -46,23 +46,54 @@ export const getOrCreateValidCartId = () => {
...
+
+    window.localStorage.setItem(storage.paypalCrUser, JSON.stringify(userStorage));
+};
+
+// Generates a random user ID.
+// This method will set the userId field and generatedUserId field.
+export const setGeneratedUserId = (id? : string, expiry? : Date) => {
+    const userStorage = getUserStorage();
+
+    const userId = id || `${ generate.generateId() }`;
```
songz: `id || generate.generateId()`

src/tracker-component.js
joshbeam: `null` is specifically for unsetting a user, per the docs

src/tracker-component.js
joshbeam: `getOrCreateValidUserId` is a convenience method that's useful in Tracker initialization

src/lib/local-storage-utils.js
```diff
@@ -76,13 +97,12 @@ export const getUserId = () => {
     return null;
 };
 
-/* Returns an existing, valid userId or creates a new one */
+/* Returns an existing, valid userId or creates a new one if it doesn't exist */
 export const getOrCreateValidUserId = () => {
     const storedValue = getUserId();
-    const now = Date.now();
```
M-OC: Did we decide to get rid of timestamps altogether?

src/tracker-component.js
```diff
@@ -234,6 +251,9 @@ export const Tracker = (config? : Config = {}) => {
         JL.tracking(trackingConfig);
     }
     const trackers = {
+        getConfig: () => {
```
M-OC: What is this for? Testing?

joshbeam: yeah, I was struggling w/ the tests because we had to use a bunch of roundabout ways to get the config, so thought I might just add a method to get it

PR #73, opened 2019-10-07
Refactor flow

M-OC:
- separates types into separate files
- simplifies types in a few places (config now refers to the user type instead of duplicating it's values)
- adds type checks into fpti


src/tracker-component.js
```diff
@@ -309,7 +308,7 @@ export const Tracker = (config? : Config = {}) => {
             createNewCartId();
             return event;
         },
-        setUser: (data : UserData) => {
+        setUser: (data : { user : UserData }) => {
```
joshbeam: Does this at all affect the fact that we can pass in `{ id }` or `{ user: id }` here?

src/types/cart.js
```diff
@@ -0,0 +1,34 @@
...
+export type CartData = {|
+    cartId? : string,
+    items : $ReadOnlyArray<Product>,
+    emailCampaignId? : string,
+    total? : string,
+    cartTotal? : string,
+    currencyCode? : string
+|};
+
+export type RemoveCartData = {|
```
joshbeam: [nit] in the spirit of refactoring, can we call this `RemoveFromCartData`?

src/tracker-component.js
M-OC: Functionally, no. But that's a good catch. Type should allow for both cases. I'll update it.

PR #77, opened 2019-10-08
refactor(property-id): change localstorage implementation

M-OC:
- separates various 'localstorage' methods into separate files. (This is responsible for most of the  file changes in this PR. Mostly updating paths in tests, adding index.js etc)
- creates new 'localstorage' utility methods for setting/getting 'propertyId'
- updates 'getPropertyId' slightly to use 'getContainer' method. This method will be utilized in a follow-up PR when getting/setting container.
- Introduces two new types: container & containerSummary


src/lib/get-property-id.js
```diff
@@ -1,40 +1,50 @@
...
+
+      return container.id;
+    })
+    .catch(() => {
+      // doing nothing for now since there's no logging
+      return '';
+    });
+};
+
+export const getContainerSettings = () : any => {
```
M-OC: This is why the "WIP" label was added.

src/lib/constants.js
```diff
@@ -5,7 +5,9 @@ export default {
   'accessTokenUrl': 'https://www.paypal.com/muse/api/partner-token',
   'storage': {
     'paypalCrUser': 'paypal-cr-user',
-    'paypalCrCart': 'paypal-cr-cart'
+    'paypalCrCart': 'paypal-cr-cart',
+    'paypalCrPropId': 'paypal-cr-propid',
+    'paypalCrContainer': 'paypal-cr-container'
```
joshbeam: what's this key for?

M-OC: It _will_ be used for container. A few pieces of this are "WIP".

src/lib/constants.js
```diff
@@ -5,7 +5,9 @@ export default {
   'accessTokenUrl': 'https://www.paypal.com/muse/api/partner-token',
   'storage': {
     'paypalCrUser': 'paypal-cr-user',
-    'paypalCrCart': 'paypal-cr-cart'
+    'paypalCrCart': 'paypal-cr-cart',
+    'paypalCrPropId': 'paypal-cr-propid',
```
songz: could we just do `propertyId` instead of `PropId`

PR #86, opened 2019-10-29
fix(1.2.0-prerelease): pageView event keys, getContainer fixes

joshbeam:
Re: 1.2.0 pre release QA testing

1. Adds external_id, shopper_id, merchant_cart_id, and product=ppshopping to page view events
2. Adds devnote about merchant provided user ID not being populated in config when it pre-exists in local storage
3. Only requests a container if merchant ID is present (in the future, we need to support querying a container by client ID as well)
4. Don't fire setUser call if no changes were made to merchant provided user info

### DIFF
--- a/src/lib/fpti.js
+++ b/src/lib/fpti.js
@@ -39,6 +39,7 @@ const resolveTrackingData = (config : Config, data : FptiInput) : any => {
   const deviceInfo = getDeviceInfo();
 
   return {
+    product: 'ppshopping',
     e: 'im',
     comp: 'ppshoppingsdk',
     page: `ppshopping:${ data.eventName }`,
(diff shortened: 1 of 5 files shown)


src/lib/fpti.js
```diff
@@ -18,7 +18,7 @@ export const sendBeacon = (src : string, data : FptiVariables | LegacyVariables)
 
   query = query.join('&');
 
-  const beaconImage = new window.Image();
+  const beaconImage = document.createElement('img');
```
songz: in case anyone is curious about this change, there's almost no difference between the 2: https://stackoverflow.com/questions/6241716/is-there-a-difference-between-new-image-and-document-createelementimg

joshbeam: Thanks @songz. I changed this so that we could use the existing image mock functions in the test suite.

PR #88, opened 2019-11-05
feat(identity): add identity to fpti events

M-OC:
- creates simple utility for creating/managing iframes.
- creates a utility for creating, interacting with the 'identity' iframe.
- creates local-storage methods for setting, getting 'identity' information.
- creates 'identity' iframe.
- adds 'identity' information to fpti events when available.
- creates webpack script for building 'identity' iframe with different environment variables.

NOTE:
Built iframe yet to be deployed to 'paypalobjects'. Url will be updated at the time this occurs.


iframes/identity/identity.js
```diff
@@ -0,0 +1,71 @@
+const env = process.env.NODE_ENV
+let targetingUrl
+
+switch (env) {
+  case 'staging':
+  case 'development':
+    targetingUrl = 'https://www.msmaster.qa.paypal.com/targeting/qraphql'
```
joshbeam: Hrm, for development I can see a use case where I want to hit a local targeting server. Maybe worth having a localhost address for development?

iframes/identity/identity.js
```diff
@@ -0,0 +1,71 @@
...
+    break;
+  case 'sandbox':
+    targetingUrl = 'https://www.sandbox.paypal.com/targeting/graphql'
+  case 'production':
+  default:
+    targetingUrl = 'https://www.paypal.com/targeting/graphql'
+    break;
+}
+
+const fetchVisitorInfo = (data) => {
```
joshbeam: [nit] More readable to destructure `(data)` -> `({ deviceInfo, country })`

iframes/identity/identity.js
```diff
@@ -0,0 +1,71 @@
...
+      if (!json || !json.data || !json.data.visitorInfo) {
+        throw new Error('response missing required fields')
+      }
+
+      return json.data.visitorInfo
+    })
+}
+
+window.addEventListener('message', async (e) => {
+  if (e.source.window !== window.parent.window) {
```
joshbeam: Can you actually compare e.source.window to window.parent.window if the domains are different?

iframes/identity/identity.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  try {
+    const visitorInfo = await fetchVisitorInfo(e.data.payload)
+    window.parent.postMessage({
+      type: 'fetch_identity_response',
+      payload: visitorInfo
+    }, '*')
+  } catch (err) {
+    //surface error in some way
```
joshbeam: Can you post a message back to the parent with an error, and let the SDK logger.error handle it?

src/lib/iframe-tools/iframe-manager.js
```diff
@@ -0,0 +1,45 @@
...
+        this.messageListeners[i](e)
+      } catch (err) {
+        console.error(err, this.messageListeners[i], e)
+      }
+    }
+  }
+
+  addMessageListener = (listener) => {
+    if (typeof listener !== 'function') {
+      console.error('listener must be a function')
```
joshbeam: Should we use `logger.error` here?

iframes/identity/identity.js
M-OC: Yeah, probably worth doing. This shouldn't be put off.

src/lib/iframe-tools/iframe-manager.js
M-OC: Good catch.

iframes/identity/identity.js
M-OC: Yeah, that'd be nice to have going forward. I'll add an additional option for local development. If a url is specified it'll be used here. If not, it'll default to msmaster.

M-OC: The only catch is that you'll still need to set up an endpoint somewhere to serve the modified iframe that points to your local instance.

src/lib/iframe-tools/iframe-manager.js
```diff
@@ -0,0 +1,45 @@
...
+  _onMessage = (e) => {
+    if (e.source.window !== this.iframe.contentWindow) {
+      return
+    }
+
+    if (e.origin !== this.url.origin) {
+      return
+    }
+
+    for (let i = 0; i < this.messageListeners.length; i++) {
```
songz: whats the reason of not using a `forEach` here?

src/tracker-component.js
```diff
@@ -219,6 +220,7 @@ export const Tracker = (config? : Config = {}) => {
   let userId;
 
   try {
+    new IdentityManager(config);
```
songz: why `new IdentityManager`? Usually I use `new` when there are possible multiple instances of the object, but I can't think of a scenario where we would want multiple `IdentityManager`

src/lib/iframe-tools/iframe-manager.js
M-OC: No reason at all. I'll update it.

iframes/identity/identity.js
```diff
@@ -0,0 +1,54 @@
...
+  if (e.data.type !== 'fetch_identity_request') {
+    return
+  }
+
+  try {
+    const visitorInfo = await fetchVisitorInfo(e.data.payload)
+    window.parent.postMessage({
+      type: 'fetch_identity_response',
+      payload: visitorInfo
+    }, '*')
```
songz: if there was a `postMessage` error, I'm afraid this may fail. Should we log an error logger?

songz: not urgent, edge case really.

test/tracker-component.test.js
```diff
@@ -485,7 +492,7 @@ describe('paypal.Tracker', () => {
     );
   });
 
-  it('should allow you to instantiate a user and then set the user', () => {
+  it.only('should allow you to instantiate a user and then set the user', () => {
```
songz: heads up, remove.

src/lib/iframe-tools/identity-manager.js
```diff
@@ -0,0 +1,65 @@
...
+    super({ src: iframeUrl });
+
+    this.addMessageListener(this.storeIdentity);
+  }
+
+  onIframeLoad = () => {
+    this.fetchIdentity();
+  }
+
+  logIframeError = (e) => {
```
songz: potential error here if `e` was undefined, or if `e.data` is null.
I think we should use lodash like `_.get`, or do `if` checks.

src/lib/iframe-tools/identity-manager.js
```diff
@@ -0,0 +1,65 @@
...
+    if (e.data.type !== 'fetch_identity_error') {
```
songz: y not log all kinds of error?

src/lib/iframe-tools/identity-manager.js
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+
+    logger.error('identity iframe error:', e.data.payload);
```
songz: usually, error objects don't have `data` key, so this may be a bit of a misnomer.
Nothing to do here, just wanted to point that out.

Screen Shot 2019-11-19 at 3 51 24 PM

PR #89, opened 2019-11-16
Limit items for setCart

insomnolent:
Rally: https://rally1.rallydev.com/#/110808537496/detail/userstory/337941446168
* We want to limit cart items to 10 or less so that users don't end up sending way too many when calling setCart
* Issues: need to set cartTotal or total but unsure if each item will always have a price...and if not then most of the test cases will fail and I'm not sure how we would be able to set an accurate cartTotal if we cannot calculate each item's price


src/lib/validation/limit-cart-items.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+import constants from '../constants';
+
+export const limitCartItems = (input : any)  => {
+  const { cartLimit } = constants;
+  if (input.items.length > 10) {
```
joshgreenwell: Should use the constant like below.

src/lib/validation/limit-cart-items.js
```diff
@@ -0,0 +1,12 @@
...
+    // eslint-disable-next-line no-console
+    console.warn('Total number of items in cart is greater than 10, cart will be truncated');
```
joshgreenwell: Same as above

src/lib/validation/legacy-input-normalizers.js
```diff
@@ -22,9 +24,9 @@ export const addToCartNormalizer = (input : any) => {
     } else if (input.cartTotal) {
       input.total = input.cartTotal;
       delete input.cartTotal;
+      input = limitCartItems(input);
```
M-OC: Just pointing out that if the user uses 'total' _and_ has more than 10 items they will never be truncated.

Talked with Song. I think it'd be ok to include this check in both cases.

insomnolent: ah okay, I'll also add it in the other case for if they have a total value then

songz: or put this after the if statement?
```
if (input) {
  ....
  input = limitCartItems(input)
}
```

src/lib/validation/limit-cart-items.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+import constants from '../constants';
+
+export const limitCartItems = (input : any)  => {
+  const { cartLimit } = constants;
+  if (input.items.length > cartLimit) {
```
songz: what if `items` is undefined? I would do this:
```
const items = input.items || []
if (items.length > cartLimit) {
  logger.....
}
input.items = items
```
Also.... we are using a `logger`, look at the other files to see how its used.

test/tracker-component.test.js
```diff
@@ -9,6 +9,286 @@ import { getUserId, getCartId } from '../src/lib/local-storage';
 
 const { sevenDays, storage } = constants;
 
+const addCartWithExtraItems = [
```
songz: Instead of creating these.... you could just have 1 items object and then do:
```
const generateItems = (input, result = []) => {
  if (result.length >== input) {
    return result
  }
  result.push({...sampleItem, id: `${sampleItem[0].id}${result.length}`})
  return generateItems(input, result)
}
...

// then do this:
const items15 = generateItems(15)
....
```

This way, you don't have to commit an array of 30 items (makes the codebase really hard to maintain, and your tests difficult to read)

src/lib/validation/limit-cart-items.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+import constants from '../constants';
+
+export const limitCartItems = (input : any)  => {
```
songz: I think it may be safer to add default value for `input` in case `null` is passed in:
```
(input = {}) => {
....
```

src/lib/validation/limit-cart-items.js
insomnolent: yep just added a check for (input.items && items.length > cartLimit)

src/lib/validation/limit-cart-items.js
```diff
@@ -0,0 +1,12 @@
...
+  const { cartLimit } = constants;
+  if (input.items && input.items.length > cartLimit) {
+    // eslint-disable-next-line no-console
+    console.warn(`Total number of items in cart is greater than ${ cartLimit }, cart will be truncated`);
```
songz: use logger instead of console

src/lib/validation/limit-cart-items.js
insomnolent: won't that not be necessary since when it's being called in legacy-input-normalizers there's already  check for if input exists?

test/tracker-component.test.js
insomnolent: ah thanks for the tip, i'll def add that instead of writing the arrays then

src/lib/validation/limit-cart-items.js
songz: make sense, thats good to know. I was afraid of some developers in the future using this incorrectly because the type is `any`.

test/tracker-component.test.js
```diff
@@ -9,6 +9,24 @@ import { getUserId, getCartId } from '../src/lib/local-storage';
 
 const { sevenDays, storage } = constants;
 
+const sampleItem = [
```
songz: `sampleItem` could just be an object, so you don't have to do `sampleItem[0]`. With arrays, the variable should be `sampleList`.

insomnolent: ah true, will change!

PR #91, opened 2019-11-26
Fix error being thrown for setCart and addCart

insomnolent:
Removed logger.error because we do not want to throw an error when truncating the cart (if there are more than 10 products). The expected behavior is for the tracker calls to still go through after limiting the cart payload

### DIFF
--- a/src/lib/validation/limit-cart-items.js
+++ b/src/lib/validation/limit-cart-items.js
@@ -1,6 +1,5 @@
 /* @flow */
 import constants from '../constants';
-import { logger } from '../logger';
 
 export const limitCartItems = (input : any)  => {
   const { cartLimit } = constants;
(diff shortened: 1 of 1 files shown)


zacharyhill:
Why would `logger.error` throw an error?

src/lib/validation/limit-cart-items.js
```diff
@@ -5,7 +5,6 @@ import { logger } from '../logger';
 export const limitCartItems = (input : any)  => {
   const { cartLimit } = constants;
   if (input.items && input.items.length > cartLimit) {
-    logger.error(`Total number of items in cart is greater than ${ cartLimit }, cart will be truncated`);
```
zacharyhill: If we're going to remove this you can remove the import as well

insomnolent: ah thanks for catching, and also good question (referring to your earlier comment - i should've clarified it in the PR description)

PR #93, opened 2019-12-10
1/3: JL refactor to fix eslint and flow errors

songz:
1. Tracking component should not be aware of JL logic. Separation of functionalities
2. JL's Tracker object lives in its own file
3. All JL logic is contained in JL file.

ESlint / flow Resources used:
* Inconsistent `this` assignment: https://eslint.org/docs/rules/consistent-this
* declare function names for stack trace: https://eslint.org/docs/rules/func-names
* function return `any` types: https://flow.org/en/docs/types/functions/#function-returns-


songz:
Merging this. Pretty safe change because the only changes to tracker (which is the main component people use) is to remove JL logic to 2 function calls (both of which probably should be wrapped around a try catch)

PR #104, opened 2020-07-24
Deprecation

songz:
A few changes

* Removes track function so the tracker component never sends out beacon again for shopping events


songz:
replaced by: https://github.com/paypal/paypal-muse-components/pull/107/files

PR #109, opened 2020-10-08
JS SDK async loading fix when using muse-components

nbierdeman:
Added belter to paypal-muse-components and used the waitForWindowReady() function instead of the DOMContentLoaded event in the setup() hook to eliminate the current race condition that had been resulting in the JS SDK failing to load async when using muse-components.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
-registry=http://registry.npmjs.org/
+registry=https://registry.npmjs.org/
 save=false
 package-lock=false
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
-registry=http://registry.npmjs.com/
+registry=http://registry.npmjs.org/
 save=false
 package-lock=false
--- a/src/component.js
+++ b/src/component.js
@@ -1,37 +1,55 @@
 /* @flow */
 
-import { getClientID, getMerchantID, getPayPalDomain, getVersion, isPayPalDomain, getEventEmitter, getDebug, getEnv, getVault, getSDKQueryParam } from '@paypal/sdk-client/src';
-import { UNKNOWN, ENV } from '@paypal/sdk-constants/src';
+import {
+  getClientID,
+  getMerchantID,
+  getPayPalDomain,
+  getVersion,
+  isPayPalDomain,
+  getEventEmitter,
+  getDebug,
+  getEnv,
+  getVault,
+  getSDKQueryParam,
+} from "@paypal/sdk-client/src";
+import { UNKNOWN, ENV } from "@paypal/sdk-constants/src";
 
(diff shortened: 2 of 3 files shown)


nbierdeman:
@bluepnume @gregjopa Ready for review at your leisure. Travis CI keeps throwing an error on build though. Hmm...

.npmrc
```diff
@@ -1,3 +1,3 @@
-registry=http://registry.npmjs.com/
+registry=http://registry.npmjs.org/
```
gregjopa, line 1: Please make this https as well
```suggestion
registry=https://registry.npmjs.org/
```

src/component.js
```diff
@@ -1,59 +1,76 @@
...
+  getDebug,
+  getEnv,
+  getVault,
+  getSDKQueryParam,
+} from "@paypal/sdk-client/src";
+import { UNKNOWN, ENV } from "@paypal/sdk-constants/src";
+
+import { logger } from "./lib/logger";
+
+import { waitForWindowReady } from "belter/dist/module";
```
gregjopa, line 19: Are we able to import this from `belter/src` instead?
```suggestion
import { waitForWindowReady } from "belter/src";
```

src/component.js
```diff
@@ -1,59 +1,76 @@
 /* @flow */
 
-import { getClientID, getMerchantID, getPayPalDomain, getVersion, isPayPalDomain, getEventEmitter, getDebug, getEnv, getVault, getSDKQueryParam } from '@paypal/sdk-client/src';
```
gregjopa, line 3: Please revert this formatting changes. That will make the PR easier to review.

src/component.js
```diff
@@ -105,23 +125,23 @@ export function insertPptm(env : string = getEnv(), isDebug : boolean = getDebug
...
+      window.paypalDDL.push({ event: "paypalButtonRender" });
     }
   });
 }
 
 export function setup() {
-  document.addEventListener('DOMContentLoaded', () => {
-    insertPptm(getEnv(), getDebug());
-  });
+  waitForWindowReady().then(() => insertPptm(getEnv(), getDebug()));
```
bluepnume, line 145: I actually think we can remove this event listener entirely. It's adding a script to `<head>` -- the window does not need to be ready to do that.

gregjopa: @bluepnume good point! We should have looked at what `insertPptm()` is doing. I agree that we don't need `waitForWindowReady` at all since it can fire right away: https://github.com/paypal/paypal-muse-components/blob/cc15563b658e258ae89b78cfb4a47c9bdb0585af/src/component.js#L104

nbierdeman: Good point! I'm headed back to the start.

PR #113, opened 2021-05-13
Add Shopping SDK component

dturgumbaev:
This PR includes changes related to new Shopping SDK.
https://engineering.paypalcorp.com/confluence/display/paypalshopping/Shopping+SDK+Events+Spike
Shopping SDk is a new component and the changes in this PR should not impact any. other muse components.
- Add base version of shopping SDK: new Shopping component, add `viewPage` event
- make sure other `paypal-muse-components` are not impacted

### DIFF
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -8,6 +8,7 @@ const initEventPublisher = (config : Config) => {
   return (converterToFpti : EventToFptiInputMapping) => {
     return (event : Object) => {
       const fptiInput : FptiInput = converterToFpti(event);
+
       trackFptiV2(config, fptiInput);
     };
   };
@@ -16,6 +17,9 @@ const initEventPublisher = (config : Config) => {
 export const setupTrackers = (config : Config) => {
   const eventPublisher = initEventPublisher(config);
   const converters = eventToFptiConverters(config);
+
   const viewPage = eventPublisher(converters.viewPageToFpti);
-  return { viewPage };
+  const productViewed = eventPublisher(converters.viewProductToFpti);
+   
+  return { viewPage, productViewed};
 };
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -1,0 +1,21 @@
+/* @flow */
+import type { FptiInput, Config } from '../types';
+
+import { trackFptiV2 } from './fpti';
+import { eventToFptiConverters, type EventToFptiInputMapping } from './shopping-event-conversions';
+
+const initEventPublisher = (config : Config) => {
+  return (converterToFpti : EventToFptiInputMapping) => {
+    return (event : Object) => {
+      const fptiInput : FptiInput = converterToFpti(event);
+      trackFptiV2(config, fptiInput);
+    };
+  };
+};
+
+export const setupTrackers = (config : Config) => {
+  const eventPublisher = initEventPublisher(config);
+  const converters = eventToFptiConverters(config);
+  const viewPage = eventPublisher(converters.viewPageToFpti);
+  return { viewPage };
+};
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -26,7 +26,7 @@ export const setupTrackers = (config : Config) => {
   const converters = eventToFptiConverters(config);
 
   const viewPage = eventPublisher(converters.viewPageToFpti);
-  const productViewed = eventPublisher(converters.viewProductToFpti);
+  const viewProduct = eventPublisher(converters.viewProductToFpti);
    
-  return { viewPage, productViewed};
+  return { viewPage, viewProduct };
 };
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -14,6 +14,13 @@ const initEventPublisher = (config : Config) => {
   };
 };
 
+/**
+ * Setup the various trackers which are a part of the shopping analytics
+ * api. The trackers setup essentially is a converter that marshalls event
+ * data in a format fpti likes.
+ * @param config
+ * @returns {{viewPage: (function(...[*]=))}}
+ */
 export const setupTrackers = (config : Config) => {
   const eventPublisher = initEventPublisher(config);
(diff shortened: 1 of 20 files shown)


src/messaging-component.js
```diff
@@ -52,7 +52,7 @@ const showExitModal = ({ cartRecovery }) => {
   return isRendered;
 };
 
-export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string, beta? : boolean } }>) => {
+export const Messaging = (...args : $ReadOnlyArray<{| cartRecovery : {| userId : string, beta? : boolean |} |}>) => {
```
zacharyhill, line 55: If someone has already integrated with this and are sending more properties in will this break their integration? Also why do we want to change this? Just to be explicit?

dturgumbaev: the change id one by running `flow-typed` I can revert them, I would need to check if this is justa. new flow syntax change

dturgumbaev: you're right the change makes it exact match https://flow.org/en/docs/types/objects
I'll revert it to make it safe

ukalbhor (approved):

src/messaging-component.js
ukalbhor: good catch

src/types/shopping-events.js
```diff
@@ -0,0 +1,19 @@
...
+export type PageType = 'HOME_PAGE' | 'CATEGORY' | 'SEARCH_RESULTS' | 'DEALS' | 'CART' | 'CHECKOUT' | 'ORDER_CONFIRMATION';
+
+export type Category = {|
+    id? : string,
+    name? : string
+|};
+
+export type PageView = {|
+    id? : string,
+    page_type? : PageType,
```
ukalbhor, line 15: `page_type` is required

src/types/shopping-events.js
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+
+import type { UserData  } from './user';
+
+export type EventType = 'pageView' | 'purchase' | 'setUser' | 'cancelCart' | 'customEvent';
+export type PageType = 'HOME_PAGE' | 'CATEGORY' | 'SEARCH_RESULTS' | 'DEALS' | 'CART' | 'CHECKOUT' | 'ORDER_CONFIRMATION';
+
+export type Category = {|
+    id? : string,
```
ukalbhor, line 9: `id` is required

ukalbhor (approved):

src/lib/shopping-analytics.js
```diff
@@ -0,0 +1,21 @@
...
+      trackFptiV2(config, fptiInput);
+    };
+  };
+};
+
+export const setupTrackers = (config : Config) => {
+  const eventPublisher = initEventPublisher(config);
+  const converters = eventToFptiConverters(config);
+  const viewPage = eventPublisher(converters.viewPageToFpti);
+  return { viewPage };
```
ukalbhor, line 20: This might be very subjective, but I found the following to be better readable 

```suggestion
  const { viewPageToFpti: viewPageToFptiConverter } = eventToFptiConverters(config);
  
  return { viewPage: initEventPublisher(config)(viewPageToFptiConverter) };
};
```

src/lib/shopping-event-conversions.js
```diff
@@ -0,0 +1,42 @@
+/* @flow */
+import type { PageView, EventType } from '../types/shopping-events';
+import type { FptiInput, Config } from '../types';
+
+import { getUserId } from './local-storage';
+
+export type EventName = "pageView" | "purchase";
```
ukalbhor, line 7: `EventName` isn't used. Instead `EventType` is used in the code

src/lib/shopping-event-conversions.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const eventToFptiConverters = (config : Config) => {
+  const eventToFptiConverter = eventToFpti(config);
+  return {
+    viewPageToFpti: (viewData : PageView) : FptiInput => {
+      return eventToFptiConverter(viewData, 'pageView');
+    }
+  };
+};
```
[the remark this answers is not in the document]
ukalbhor: While the higher order functions is a good pattern and facilitates proper use of `config`, I found it generally tedious to follow the logic. 


Here is a simple alternative that doesn't use HOF 

```javascript

const getDefaultFPTIEvent = () => {
  const data : FptiInput = {
    shopperId : getUserId().userId,
    merchantProvidedUserId : getUserId().merchantProvidedUserId,
  };

  return data;
}

function getDefaultFPTIEventFromShoppingEvent(config : Config,
                                              event : Object,
                                              eventType : EventType) : FptiInput {
  if (!event.user) {
    event.user = config.user;
  }

  const data : FptiInput = {
    eventName: eventType,
    eventType,
    eventData: JSON.stringify(event)
  };
}

/**
 * Generic function that performs the default conversion for events
 * by stringifying the event object
 *
 * @param config
 * @param event
 * @param eventType
 * @returns {FptiInput}
 */
const getDefaultConversion = (config: Config,
                              event : Object,
                              eventType : EventType) => {
  const data : FptiInput =
      Object.assign(
          getDefaultFPTIEvent(),
          getDefaultFPTIEventFromShoppingEvent(config, event, eventType));

  return data;
}


export const eventToFptiConverters = (config : Config) => {
  return {
    viewPageToFpti: (viewData : PageView) : FptiInput => {
      return getDefaultConversion(config, viewData, 'pageView');
    }
  };
};


```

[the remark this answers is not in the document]
dturgumbaev: thanks, I agree I like it this way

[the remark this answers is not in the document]
dturgumbaev: i made more of a hybrid approach. I like the idea of splitting but the construction of FPTI input should be more of atomic method. So, I have refactored a bit your proposal

zacharyhill (approved):

src/types/shopping-events.js
dturgumbaev: I'll update the doc, to make it optional. I'd rather keep it optional at 1st release

PR #114, opened 2021-05-18
Add additional test for pageView HOME_PAGE

ukalbhor:
* Add comments to shopping-event types b3bb71e
* Add unicode characters to shopping event conversion tests 88c79b0 
* Add test for pageView HOME_PAGE ada9d84

### DIFF
--- a/src/lib/shopping-event-conversions.js
+++ b/src/lib/shopping-event-conversions.js
@@ -10,50 +10,93 @@ import { getUserId } from './local-storage';
 
 export type EventToFptiInputMapping = (event : Object) => FptiInput;
 
-function getStoredUserIds() : Object {
+const getDefaultFPTIEvent = () => {
   const storedUserIds = getUserId();
-  if (storedUserIds) {
-    return {
-      shopperId: storedUserIds.userId,
-      merchantProvidedUserId: storedUserIds.merchantProvidedUserId
-    };
-  } else {
-    return {};
-  }
-}
 
-function convertShoppingEventToFptiInput(
-  config : Config,
-  event : Object,
-  eventType : EventType
-) : FptiInput {
-  if (!event.user) {
-    event.user = config.user;
-  }
+  const data = {
+    shopperId: storedUserIds ? storedUserIds.userId : undefined,
+    merchantProvidedUserId: storedUserIds ? storedUserIds.merchantProvidedUserId : undefined
+  };
 
-  const storedUserIds = getStoredUserIds();
+  return data;
+};
+
+function getDefaultFPTIEventFromShoppingEvent(config : Config,
+  shoppingEvent : Object,
+  eventType : EventType) : FptiInput {
+
+  const eventData = {
+    ...shoppingEvent,
+    user: shoppingEvent.user || config.user
+  };
 
   const data : FptiInput = {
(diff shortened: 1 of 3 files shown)


src/lib/shopping-event-conversions.js
```diff
@@ -10,50 +11,93 @@ import { getUserId } from './local-storage';
 
 export type EventToFptiInputMapping = (event : Object) => FptiInput;
 
-function getStoredUserIds() : Object {
+const getDefaultFPTIEvent = () => {
```
dturgumbaev, line 14: the name is confusing, it is not a FPTI event, we extracting some attributes. That will be used in construction of F{TI event.

dturgumbaev: maybe `getDefaultFPTIEventAttributes`. The wording default is very obscure, I know you trying to convey a generic nature of the function, but without reading and delving into a code it is hard to understand. Maybe `fetchFPTIAttributesFromUserContext`

dturgumbaev:
Did you run `npm run test` it will execute flow-typed validation

src/lib/shopping-event-conversions.js
```diff
@@ -10,50 +11,93 @@ import { getUserId } from './local-storage';
...
+      Object.assign(
+        getDefaultFPTIEvent(),
+        { ...defaultConvertedEvent,
+          eventData: JSON.stringify(defaultConvertedEvent.eventData) }
+      );
+
+  return data;
+};
+
+const getViewProductEventConvertor = (config : Config,
```
dturgumbaev, line 69: it is a bit harder to read. all we're doing is update the payload with currency if one defined in config. you can create a copy of ProductView event with currency updated and then call getDefaultConvertor.

src/lib/shopping-event-conversions.js
```diff
@@ -10,50 +11,93 @@ import { getUserId } from './local-storage';
...
-  }
+  const data = {
+    shopperId: storedUserIds ? storedUserIds.userId : undefined,
+    merchantProvidedUserId: storedUserIds ? storedUserIds.merchantProvidedUserId : undefined
+  };
+
+  return data;
+};
+
+function getDefaultFPTIEventFromShoppingEvent(config : Config,
```
dturgumbaev, line 25: let's just call it `getFPTIEventFromShoppingEventConverter`, again not a big fan of generic names `generic`, `default`, `info`. I do not think they bring any clarity

src/lib/shopping-event-conversions.js
```diff
@@ -10,50 +11,93 @@ import { getUserId } from './local-storage';
...
+ * by stringifying the event object
+ *
+ * @param config
+ * @param shoppingEvent
+ * @param eventType
+ * @returns {FptiInput}
+ */
+const getDefaultConvertor = (config : Config,
+  shoppingEvent : Object,
+  eventType : EventType) => {
```
dturgumbaev, line 54: add the the return type to function signature is `:FPTIInput`

dturgumbaev:
the change otherwise, looks good. Please, fix the build. Make sure to run `npm run test` locally

dturgumbaev (approved):

PR #115, opened 2021-05-20
Implement control mechanism to publish FPTI events

dturgumbaev:
The FPTI Shopping analytics events may only be published if merchant has XO container.

### DIFF
--- a/src/lib/shopping-fpti-event-publisher.js
+++ b/src/lib/shopping-fpti-event-publisher.js
@@ -1,0 +1,90 @@
+/* @flow */
+import type { FptiInput, Config, ContainerSummary } from '../types';
+
+import { trackFptiV2 } from './fpti';
+import { fetchContainerSettings } from './get-property-id';
+import { logger } from './logger';
+
+/**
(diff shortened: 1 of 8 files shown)


zacharyhill (approved):

src/lib/shopping-fpti-event-publisher.js
```diff
@@ -0,0 +1,55 @@
...
+export const ShoppingEventPublisher = (config : Config) => {
+
+  const trackEventQueue = [];
+
+  function isAllowedToPublishEvent() : boolean {
+    const isContainerIdSet  = (config.propertyId !== undefined && config.propertyId !== null && config.propertyId.length > 0);
+    return isContainerIdSet;
+  }
+
+  function publishFptiEvent(fptiEvent : FptiInput) {
```
ukalbhor, line 17: minor : can you add a short comment on why some events are okay to be published immediately and some are queued ? - that we need to wait until the containerSummary is returned and that this queuing is done only at the beginning.

ukalbhor (approved):

PR #119, opened 2021-05-25
Automagically generate product view payload from open graph tags

ukalbhor:
DTSHOPSDK-172

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -1,0 +1,31 @@
+/* @flow */
+
+const autoGenerateProductPayload = () => {
+  const tags = {
+    'og:title': 'product_name',
+    'product:price:amount': 'price',
+    'product:price:currency': 'currency',
+    'og:description': 'description',
+    'og:url': 'url'
+  };
+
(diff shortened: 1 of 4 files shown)


zacharyhill (approved):

src/lib/utils.js
```diff
@@ -0,0 +1,31 @@
...
+const autoGenerateProductPayload = () => {
+  const tags = {
+    'og:title': 'product_name',
+    'product:price:amount': 'price',
+    'product:price:currency': 'currency',
+    'og:description': 'description',
+    'og:url': 'url'
+  };
+
+  const attributes = Object.entries(tags).map(([ k, v ]) => {
```
batiskafff, line 12: Variable names **k** and **v** is not self documented. It looks like it stands from **key** and **value**, but it isn't explain business value

dturgumbaev (approved):

src/lib/utils.js
ukalbhor: ok, I've made it more clearer.

PR #120, opened 2021-06-07
Introduce new setUser function in the shopping analytics API

ukalbhor:
Review and merge after https://github.com/paypal/paypal-muse-components/pull/119


src/lib/utils.js
```diff
@@ -0,0 +1,31 @@
+/* @flow */
+
+const autoGenerateProductPayload = () => {
```
dturgumbaev, line 3: this snippet looks like from different PR

dturgumbaev (approved):

src/lib/utils.js
ukalbhor: yes, they have the same commits as the other .... once the other is merged these commits will not be shown here

PR #121, opened 2021-06-16
Implement generic event send() method

dturgumbaev:
- Update ShoppingAnalytics with new method to publish events. It should take an event name and payload. 
- Disable container verification for integrating with SDK until the lookup by client id is added
- loosen up the structure of the existing events.

### DIFF
--- a/src/lib/shopping-fpti-event-publisher.js
+++ b/src/lib/shopping-fpti-event-publisher.js
@@ -25,11 +25,12 @@ export const ShoppingEventPublisher = (config : Config) => {
    * @returns {boolean}
    */
   function isAllowedToPublishEvent() : boolean {
-    const isContainerIdSet =
-      config.propertyId !== undefined &&
-      config.propertyId !== null &&
-      config.propertyId.length > 0;
-    return isContainerIdSet;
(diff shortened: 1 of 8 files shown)


zacharyhill (approved):

src/lib/shopping-fpti-event-publisher.js
```diff
@@ -25,11 +25,12 @@ export const ShoppingEventPublisher = (config : Config) => {
...
-      config.propertyId !== undefined &&
-      config.propertyId !== null &&
-      config.propertyId.length > 0;
-    return isContainerIdSet;
+    // const isContainerIdSet =
+    //   config.propertyId !== undefined &&
+    //   config.propertyId !== null &&
+    //   config.propertyId.length > 0;
+    // return isContainerIdSet;
+    return true;
```
batiskafff, line 33: Is it debug code?

dturgumbaev: As mentioned in the comments, we need to disable  the check for container to control FPTI event publishing. The idea here is we will need to update TMNW Api to include fetch container by client id.

PR #122, opened 2021-06-21
Exclude from loading the muse component in case of shopping

dturgumbaev:
Currently, PPTM.js gets loaded automatically for any PP SDK component. We want to make an exception for Shopping component as shopping component will be a replacement of PPTM.js and merchants will include it every page on thier website. This may interfere with store cash events.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -7,13 +7,16 @@ import { logger } from './lib/logger';
 
 export const PPTM_ID = 'xo-pptm';
 
+const SHOPPING_COMPONENT = 'shopping';
+
+
 /*
 Generates a URL for pptm.js, e.g. http://localhost:8001/tagmanager/pptm.js?id=www.merchant-site.com&t=xo&mrid=xyz&client_id=abc
 */
 export function getPptmScriptSrc(paypalDomain : string, mrid : ?string, clientId : ?string, url : string) : string {
   // "xo" is a checkout container
   const type = 'xo';
-
+  
   // We send this so that we know what version of the Payments SDK the request originated from.
   const version = getVersion();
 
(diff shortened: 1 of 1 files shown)


src/component.js
```diff
@@ -67,6 +67,12 @@ function _isPayPalDomain() : boolean {
   return window.mockDomain === 'mock://www.paypal.com' || isPayPalDomain();
 }
 
+function _isShoppingSdkComponent() : boolean {
+  const componentParam = (getSDKQueryParam('components') || '').toLowerCase();
```
bluepnume, line 71: I'd recommend using `getComponents()` and doing an `indexOf()` check rather than just checking if the entire param === 'shopping'.

https://github.com/paypal/paypal-sdk-client/blob/6b4fe0dd9d31f4176980a007a908864328f834b0/src/global.js#L93

Can we also use the `COMPONENTS` enum here rather than hard-coding?

dturgumbaev: Thanks, I'll update the PR

src/component.js
```diff
@@ -75,7 +90,12 @@ export function insertPptm(env : string = getEnv(), isDebug : boolean = getDebug
     // website, and then it'll render an iframe from the PayPal domain which will in turn
     // initialize the SDK again. We don't want to insert another pptm.js on the paypal.com
     // domain, though.
-    if (!_isPayPalDomain()) {
+
+    // We do not want to load pptm.js for shopping component. Merchant may always add muse, checkout button  to load pptms.js if neeeded.
+    // However, shopping component will be a replacement of pptm.js. Some merchant may use it to publish store cash events.
+    // https://engineering.paypalcorp.com/jira/browse/DTSHOPSDK-295
+    
+    if ((!_isPayPalDomain()) && (!_isShoppingSdkComponent())) {
```
zacharyhill, line 98: Should this not be an `||`? As an `&&` the original conditional `(!_isPayPalDomain())` will no longer act as it did before.

dturgumbaev: it is negation that makes it a bit confusing. What we trying to do here is: 
either `_isPayPalDomain()` || `_isShoppingSdkComponent()` => do nothing
or 
`!_isPayPalDomain()` && `!_isShoppingSdkComponent()` => load the muse
so the logic is fine here

zacharyhill (approved):

PR #123, opened 2021-06-22
Add set() to update set shopping attributes

dturgumbaev:
Provide a new method in SDK - set(), to set any shopping variables that will be included  in the events. 
Currently, these settings are per page context.

### DIFF
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -9,6 +9,8 @@ import {
   type EventToFptiInputMapping
 } from './shopping-event-conversions';
 import { setUser } from './user-configuration';
+import { shoppingAttributes } from './shopping-attributes';
+
 
 const initEventPublisher = (config : Config) => {
   const fptiEventPubisher = ShoppingEventPublisher(config);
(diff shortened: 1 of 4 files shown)


zacharyhill:
I'm a little confused -- what will this be used for?

dturgumbaev:
> I'm a little confused -- what will this be used for?

Good question, I will update the Confluence page. The idea here, SDK will provide a set() for merchant to set any attributes, that will be populated in every event. Attributes could be:
'currency', 'country' etc. We can even pass some user specific attributes: 'high_value_ccustomer' etc
This is similar to Google Analytics
https://developers.google.com/analytics/devguides/collection/ga4/persistent-values

zacharyhill (approved):
@dturgumbaev thanks for the explanation. Lgtm.

src/lib/shopping-analytics.js
```diff
@@ -25,6 +27,8 @@ function initGenericEventPublisher(config : Config) : Object {
   const convertEvent = eventToFptiConverters(config).eventToFpti;
   return {
     publishEvent: (event : EventType, payload : Object) => {
+      const prevPayload = payload || {};
+      payload = { ...prevPayload, ...config.shoppingAttributes };
```
batiskafff, line 31: Overriding payload component you modifying original `payload` object. Object is passing here `as a reference`. 
I don't think it is your intentional behavior.

It should work fine in this case
```
const extendedPayload = {...payload, ...config.shoppingAttributes}
```

dturgumbaev: I agree, ideally object should be immutable. I'll make a change

PR #124, opened 2021-06-28
Set user v2

zacharyhill:
**JIRA:**
DTSHOPSDK-366

**Changes:**
Changed overall way we expect merchants to set known user ID. Did this by removing `setUser` method and instead expecting `set` method to be called with `user_id` property.

**Example 1 (no user ID set by merchant):**

Merchant calls:

```
const t = PayPalShoppingSDK.ShoppingAnalytics();
t.send('page_view', { id: 'HOME' });
```

Sample of relevant merchant properties:

```
sinfo: "{\"id\":\"HOME\"}"
shopper_id: "uid_6f9a2e15eb_mtc6mzq6mza"
product: "ppshopping_v2"
external_id: undefined
```

**Example 2 (user ID set by merchant):**

Merchant calls:

```
const t = PayPalShoppingSDK.ShoppingAnalytics();
t.set({ user_id: '123' });
t.send('page_view', { id: 'HOME' });
```

Sample of relevant merchant properties:

```
sinfo: "{\"id\":\"HOME\"}"
shopper_id: "uid_6f9a2e15eb_mtc6mzq6mza"
product: "ppshopping_v2"
external_id: "123"
```

### DIFF
--- a/src/lib/shopping-event-conversions.js
+++ b/src/lib/shopping-event-conversions.js
@@ -28,18 +28,17 @@ function convertShoppingEventToFptiInput(
   eventType : EventType
 ) : FptiInput {
 
-  if (!event.user) {
-    event.user = config.user;
-  }
-
   const storedUserIds = getStoredUserIds();
 
+  const eventDataPayload = { ...event };
+  // remove user_id property from event_payload since it is passed separately
+  delete eventDataPayload.user_id;
   const data : FptiInput = {
     eventName: eventType,
     eventType,
-    eventData: JSON.stringify(event),
+    eventData: JSON.stringify(eventDataPayload),
     shopperId: storedUserIds.shopperId,
-    merchantProvidedUserId: storedUserIds.merchantProvidedUserId
+    merchantProvidedUserId: event.user_id
   };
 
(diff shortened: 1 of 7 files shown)


src/lib/shopping-event-conversions.js
```diff
@@ -28,18 +28,17 @@ function convertShoppingEventToFptiInput(
   eventType : EventType
 ) : FptiInput {
 
-  if (!event.user) {
-    event.user = config.user;
-  }
-
   const storedUserIds = getStoredUserIds();
 
+  const eventDataPayload = { ...event };
```
dturgumbaev, line 33: why we need to do this, Please, share examples

zacharyhill: I'm doing this for the next part where I remove `user_id` from the payload so it is not repeated in `sinfo`.

https://github.com/paypal/paypal-muse-components/pull/124/files#diff-eacd74ffe70966f9d40b06ba2c6584ad610922a4cde7772f73d360f39f75469cR34

dturgumbaev (approved):

PR #127, opened 2021-07-12
Introduce event specific processors

dturgumbaev:
- Add event specific attributes
- refactor FPTI event input mapping: the event to FPTInput is done in one place. This constructs FPTIInput from device information, user identity and event to FPTI mapping
- Deprecate productView


src/lib/shopping-fpti/event-handlers.js
```diff
@@ -0,0 +1,49 @@
...
+      return {
+        es: 'visitorInfoFlowStarted',
+        fltp: 'store-cash',
+        offer_id: offerId
+      };
+    }
+    return {};
+  }
+
+  function enrichPurchaseEvent() : Object {
```
batiskafff, line 18: Why does it named **enrich**PurchaseEvent ?

src/types/fpti.js
```diff
@@ -53,7 +53,13 @@ export type FptiInput = {|
   // SDK generated user ID
   shopperId? : string,
   // Merchant provided user ID
-  merchantProvidedUserId? : string
+  merchantProvidedUserId? : string,
+  
+  es? : string,
+
+  fltp? : string,
```
batiskafff, line 60: Isn't it mandatory field?

test/lib/fpti-test-utils.js
```diff
@@ -0,0 +1,74 @@
...
+
+export function verifyFptiIdentityMappings(fptiEvent, identityData) {
+  expect(fptiEvent.confidenceScore).toEqual(identityData.confidenceScore);
+  expect(fptiEvent.encryptedAccountNumber).toEqual(
+    identityData.encryptedAccountNumber
+  );
+  expect(fptiEvent.identificationType).toEqual(identityData.identificationType);
+}
+
+export function verifyFptiDeviceInfoMappings(fptiEvent, deviceInfoInput) {
```
batiskafff, line 57: Why didn't you use following assert?
```
expect(fptiEvent).toEqual(deviceInfoInput);
```

src/lib/shopping-fpti/event-handlers.js
dturgumbaev: it return new FPTI attributes based on the event type

src/types/fpti.js
dturgumbaev: nope it is not

test/lib/fpti-test-utils.js
dturgumbaev: only the subset of fptiEvent attributes are mapped from deviceInfoInput

batiskafff (approved):

src/lib/shopping-fpti/event-handlers.js
```diff
@@ -0,0 +1,104 @@
+/* @flow */
+import type { Config } from '../../types';
+import type { EventType } from '../../types/shopping-events';
+
+export function eventSifnoBuilderInit(config : Config) : Object {
```
zacharyhill, line 5: Can you fix this to be `eventSinfoBuilderInit` -- looks like `fn` `should be `nf`. Not a big deal but might avoid some confusion in the future.

zacharyhill (approved):

src/lib/shopping-fpti/event-handlers.js
dturgumbaev: I'll fix it before merge

PR #128, opened 2021-07-16
Dtshopsdk 391

batiskafff:
* Add webpack configuration, now you can run identity on local environment
* Add fetch country to identity iframe
* Store Country to local storage
* Add constants for better cross file handling
* Move identity fetch to Promise instead of callback model

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,6 @@
 [ignore]
 .*/node_modules/babel-plugin-flow-runtime
+.*/node_modules/@paypal/sdk-client
 .*/node_modules/npm
 .*/node_modules/jsonlint
 .*/dist/module
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,6 +1,5 @@
 [ignore]
 .*/node_modules/babel-plugin-flow-runtime
-.*/node_modules/@paypal/sdk-client
 .*/node_modules/npm
 .*/node_modules/jsonlint
 .*/dist/module
@@ -14,6 +13,7 @@
 .*/test/local-storage/local-storage-utils.test.js
 .*/test/tracker-component.test.js
 .*/test/.*
+<PROJECT_ROOT>/webpack.dev.mjs
 [untyped]
 .*/src/lib/jetlore
 .*/src/storeCash.js
--- a/iframes/identity/userCountry.js
+++ b/iframes/identity/userCountry.js
@@ -1,0 +1,28 @@
+export const fetchUserCountry = () => {
+  const US = 'US';
+  const geoLocationUrl = 'https://www.paypalobjects.com/muse/noop.js';
+
+  return fetch(geoLocationUrl, { mode: 'no-cors' })
+    .then(res => {
+      // we need this function to workaround a chrome issue
+      // https://stackoverflow.com/questions/45816743/how-to-solve-this-caution-request-is-not-finished-yet-in-chrome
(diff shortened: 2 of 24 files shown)


test/lib/user-configuration.test.js
```diff
@@ -20,42 +19,43 @@ describe('sets up user details', () => {
...
-  it('should set up user details for configuration with user details', () => {
-    const config = { user: { id: 'test_user' } };
-    setupUserDetails(config);
-    expect(IdentityManager).toBeCalledWith(config, undefined);
-    expect(getOrCreateValidUserId).toHaveBeenCalledTimes(1);
-    expect(config.user.id).toEqual(mockedUserId);
-    expect(setMerchantProvidedUserId).toBeCalledWith('test_user');
-    expect(config.user.merchantProvidedUserId).toEqual('test_user');
-  });
+  // it('should set up user details for empty configuration', () => {
```
batiskafff, line 22: Check this

.flowconfig
```diff
@@ -13,6 +13,7 @@
 .*/test/local-storage/local-storage-utils.test.js
 .*/test/tracker-component.test.js
 .*/test/.*
+<PROJECT_ROOT>/webpack.dev.mjs
```
dturgumbaev, line 16: why we need this?

iframes/identity/userCountry.js
```diff
@@ -0,0 +1,28 @@
+export const fetchUserCountry = () => {
+  const US = 'US';
+  const geoLocationUrl = 'https://www.paypalobjects.com/muse/noop.js';
+
+  return fetch(geoLocationUrl, { mode: 'no-cors' })
```
dturgumbaev, line 5: is fetch a country call goes via iframe? if so do we still need no-cors parameter

src/messaging-component.js
```diff
@@ -52,7 +52,7 @@ const showExitModal = ({ cartRecovery }) => {
   return isRendered;
 };
 
-export const Messaging = (...args : $ReadOnlyArray<{ cartRecovery : { userId : string, beta? : boolean } }>) => {
+export const Messaging = (...args : $ReadOnlyArray<{| cartRecovery : {| userId : string, beta? : boolean |} |}>) => {
```
dturgumbaev, line 55: this is updated by flow. I believe it sets strict definition of the type. I usually remove this change, as I do not mean to change existing classes

src/types/user.js
```diff
@@ -2,12 +2,23 @@
 export type UserData = {|
   merchantProvidedUserId? : string,
   id? : string,
-  email? : string,
+  email : string,
```
dturgumbaev, line 5: email should be optional

.flowconfig
batiskafff: Ignore type checking in webpack configuration

iframes/identity/userCountry.js
batiskafff: Yes we do. Iframe and noop .js live on different domains. `paypal.com` vs `paypalobjects.com`

src/types/user.js
batiskafff: fixed

src/messaging-component.js
batiskafff: fixed

dturgumbaev (approved):

PR #130, opened 2021-07-26
disable offer setting

dturgumbaev:
add a toggle to disable offer attributes

### DIFF
--- a/src/lib/shopping-fpti/event-handlers.js
+++ b/src/lib/shopping-fpti/event-handlers.js
@@ -2,9 +2,14 @@
 import type { Config } from '../../types';
 import type { EventType } from '../../types/shopping-events';
 
+function findConfigurationAttribute(config : Config, payload : Object = {}, attribName : string) : ?string {
+  const shopperConfig = config.shoppingAttributes || {};
+  return shopperConfig[attribName] || payload[attribName];
+}
+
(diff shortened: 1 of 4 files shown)


src/lib/shopping-fpti/event-handlers.js
```diff
@@ -32,10 +37,11 @@ export function eventSinfoBuilderInit(config : Config) : Object {
 }
 
 export function customEventMappingInit(config : Config) : Object {
-  function enrichPageViewEvent() : Object {
+  function enrichPageViewEvent(payload : Object) : Object {
+    const disableStoreCash = findConfigurationAttribute(config, payload, 'disable_storecash') || 'false';
     const containerSummary = config.containerSummary || {};
     const offerId = containerSummary.programId;
-    if (offerId) {
+    if ((disableStoreCash === 'false') && offerId) {
```
zacharyhill, line 44: Is there a reason we need to use a string for `disable_storecash` property?

zacharyhill (approved):

src/lib/shopping-fpti/event-handlers.js
dturgumbaev: Yes, we need to be able to have an option to disable store cash on any page

PR #138, opened 2021-09-17
Add generatePageAutoData for parsing ldjson and og tags

ukalbhor:
53a8e79 fix issues with flow
7ab74d1 fix eslint suggestions
397eb0f support deeply nested itemprop DOM nodes
fcee9b0 special handling for 'item' microdata tags
20df52a include microdata json in the event data we capture
51074c4 remove caching for captured data as it won't work correct for SPAs
4ae4131 add parser for microdata tags on the page
5334743 add generatePageAutoData for parsing ldjson and og tags
116345d add _get helper method


src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,42 @@
...
+        autoData.push(ogTagsData);
+    }
+    const ldTagsData = generateJSONldData();
+    if (ldTagsData) {
+        autoData.push(ldTagsData);
+    }
+
+    window.__pp__shopping__ = window.__pp__shopping__ || {};
+    window.__pp__shopping__.parsed_data = autoData;
+}
```
ukalbhor, line 42: TODO - Add tests

batiskafff (approved):

src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,43 @@
+import parseOgTags from './og-parser';
+import parseJsonLd from './jsonld-parser';
+import { _get } from "../utils";
+
+const generateOgData = () => {
+    const ogTags = parseOgTags();
```
denkt315, line 6: I don't see explicit documentation of format that is produced by this parser. What parser is doing is assembling tree structure objects and building some kind of graph. Unfortunately this graph is not standardized so it will be hard to consume on backed. Can we stick to original format  key: value

src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,43 @@
...
+    const hasOgTags = Array.isArray(ogTags) && ogTags.length > 0
+    return hasOgTags ? {
+        type: "open_graph",
+        version: 1,
+        data: ogTags
+    } : null;
+}
+
+const generateJSONldData = () => {
+    const ldTags = parseJsonLd();
```
denkt315, line 16: Please keep in mind that we need to send/parse only limited number of json ld structures. https://schema.org/Product and https://schema.org/BreadcrumbList

src/lib/shopping-analytics.js
```diff
@@ -44,6 +45,7 @@ export const shoppingAnalyticsSetup = (config? : Config = {}) => {
     viewPage: shoppingTracker.viewPage,
     send: sendOrEnqueue,
     set: shoppingTracker.set,
-    autoGenerateProductPayload: shoppingTracker.autoGenerateProductPayload
+    autoGenerateProductPayload: shoppingTracker.autoGenerateProductPayload,
```
denkt315, line 48: We don't need this function. I'm not sure that anybody is using this

src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,43 @@
...
+    if (ogTagsData) {
+        autoData.push(ogTagsData);
+    }
+    const ldTagsData = generateJSONldData();
+    if (ldTagsData) {
+        autoData.push(ldTagsData);
+    }
+
+    window.__pp__shopping__ = window.__pp__shopping__ || {};
+    window.__pp__shopping__.parsed_data = autoData;
```
denkt315, line 42: Implementation fro product (that we had before) was returning this.  Why here we use global variable `window.__pp__shopping__.parsed_data = autoData;`

denkt315 (changes requested):

src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,43 @@
+import parseOgTags from './og-parser';
```
ukalbhor, line 1: `og:` is not the best regex to parse open graph tags

src/lib/tag-parsers/generate-page-auto-data.js
ukalbhor: Has to be limited since the kafka message has a max payload size.

src/lib/tag-parsers/generate-page-auto-data.js
```diff
@@ -0,0 +1,43 @@
...
+        type: "schema.org/ld+json",
+        version: 1,
+        data: ldTags
+    }  : null;
+}
+
+export const generatePageAutoData = () => {
+    if (Array.isArray(_get(window, "__pp__shopping__.parsed_data"))) {
+        return window.__pp__shopping__.parsed_data;
+    }
```
ukalbhor, line 28: How do we handle caching when SPA ?

src/lib/tag-parsers/generate-page-auto-data.js
ukalbhor: for now we only care about product and breadcrumbs. So this implementation works.

src/lib/utils.js
```diff
@@ -47,11 +47,19 @@ export const autoGenerateProductPayload = () : ?GeneratedProductPayload => {
 };
 
 export const tryAndLog = (fn : Function) => {
-  return () => {
+  return (argObj) => {
     try {
-      return fn();
+      return fn(argObj);
     } catch (err) {
       console.log(err); /* eslint-disable-line no-console */
```
dturgumbaev, line 54: lets not log it to console

dturgumbaev (approved):
please, fix the build issues before merge

PR #139, opened 2021-09-22
Skip identity iframe fetch for identified user

dturgumbaev:
Skip identity iframe fetch for identified user


src/lib/user-configuration.js
```diff
@@ -17,10 +18,25 @@ import { logger } from './logger';
...
+  callback : Function
+) {
+  const identity = getIdentity();
+  if (identity) {
+    callback(identity, null);
+  } else {
+    fetchUserIdentity(config, callback);
+  }
+}
+
```
ukalbhor, line 39: Can you double check if we need this piece of code please ? -- https://github.com/paypal/paypal-muse-components/blob/master/src/lib/iframe-tools/identity-manager.js#L59-L79

Or why that piece of code isn't working as expected and returning a cached response.

src/lib/user-configuration.js
```diff
@@ -17,10 +18,25 @@ import { logger } from './logger';
...
+  callback : Function
+) : IdentityManager {
   return new IdentityManager(config, callback);
 }
 
+function setupUserIdentity(
+  config : Config,
+  callback : Function
+) {
+  const identity = getIdentity();
```
ukalbhor, line 32:
```suggestion
  const cachedIdentity = getIdentity();
```

src/lib/user-configuration.js
ukalbhor: Further, here we do not load the iframe since we do not invoke `IdentityManager` which hands back control to the iframe. The iframe then sends out a tracking event (as is the case in your latest PR in MNW). 

Just want to double check if the track event is still being sent.

dturgumbaev: This is OK, we may actually remove it next time. All it does it load iframe and then checks if ideintity and if missing starts targeting fetch. 
This PR is to skip loading iframe all together. 
https://github.com/paypal/paypal-muse-components/blob/master/src/lib/iframe-tools/identity-manager.js#L59-L79

dturgumbaev: The IdentityManager is purely for visitor fetch (in that sense it i different from PPTM where bundle flow is linked with Iframe), it has no functionality to publish events. Currently, all events are published from main SDK code. 
https://github.com/paypal/paypal-muse-components/blob/master/iframes/identity/identity.js

ukalbhor: ah, I understand

batiskafff (approved):

PR #148, opened 2021-11-02
DTSHOPSDK-675: Expose 'getPageSKUData method.

batiskafff:
-  Expose 'getPageSKUData method
-  Fix typo
-  Remove unnecessary 'Fix me'

### DIFF
--- a/src/lib/shopping-analytics.js
+++ b/src/lib/shopping-analytics.js
@@ -6,6 +6,8 @@ import type { Config } from '../types/config';
 import { setupTrackers } from './shopping-trackers';
 import { setupUserDetails } from './user-configuration';
 import { setupContainer } from './get-property-id';
+// $FlowFixMe
+import { capturePageData } from './tag-parsers/capture-page-data';
 
 // $FlowFixMe
 export const shoppingAnalyticsSetup = (config? : Config = {}) => {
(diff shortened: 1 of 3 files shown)


src/lib/shopping-analytics.js
```diff
@@ -66,6 +66,7 @@ export const shoppingAnalyticsSetup = (config? : Config = {}) => {
     send: sendOrEnqueue,
     set: shoppingTracker.set,
     autoGenerateProductPayload: shoppingTracker.autoGenerateProductPayload,
+    getPageSkuData: capturePageData,
```
ukalbhor, line 69: Should this method somehow indicate it isn't for public use ? 
Eg: `__getPageSkuData` or `private__getPageSkuData` ? 

The idea is the capturing of SKU data would be transparent to the end user and will be controlled by a config.

batiskafff: Is any other methods except `set` and `send` designed for public use?

ukalbhor: I think we can also remove `autoGenerateProductPayload` .... it isn't used on live .... @dturgumbaev - can you confirm ?

dturgumbaev (approved):

PR #150, opened 2021-11-03
Fail tests when adding new props to fpti

zacharyhill:
DTSHOPSDK-610

**Purpose of this PR:**
We wanted to ensure our docs stay updated when new FPTI properties are added to events. In order to do this, we are checking included properties and supplying to link to update docs when the tests fail.

**What it will look like when any of these tests fail:**
image

### DIFF
--- a/src/lib/shopping-fpti/event-handlers.js
+++ b/src/lib/shopping-fpti/event-handlers.js
@@ -8,7 +8,7 @@ function findConfigurationAttribute(config : Config, payload : Object = {}, attr
   return shopperConfig[attribName] || payload[attribName];
 }
 
-export const includedAttributes = [
+export const allowedAttributes = [
   // page view
   'page_type',
   'page_name',
@@ -38,7 +38,7 @@ export const includedAttributes = [
 export function eventSinfoBuilderInit(config : Config) : Object {
   function filterAttributesForSinfoPayload(event : Object) : Object {
     const filteredAttributes = Object.keys(event)
-      .filter((key) => includedAttributes.includes(key))
+      .filter((key) => allowedAttributes.includes(key))
       .reduce((obj, key) => {
         obj[key] = event[key];
         return obj;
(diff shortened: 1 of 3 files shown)


src/lib/shopping-fpti/event-handlers.js
```diff
@@ -8,16 +8,41 @@ function findConfigurationAttribute(config : Config, payload : Object = {}, attr
   return shopperConfig[attribName] || payload[attribName];
 }
 
+export const includedAttributes = [
```
ukalbhor, line 11: minor : maybe `allowedAttributes` ?

src/lib/shopping-fpti/event-handlers.js
```diff
@@ -8,16 +8,41 @@ function findConfigurationAttribute(config : Config, payload : Object = {}, attr
...
+  'product_price',
+  'product_brand',
+  'product_category_name',
+  'product_category_id',
+  'product_discount',
+  // purchase
+  'amount',
+  // set properties
+  'currency',
+];
```
ukalbhor, line 36: This is totally fine .... but this seems like the perfect fit for a `@flow` type where we get this check for 'free'.

zacharyhill: Can we use the flow type and not affect the functionality of the code? Is flow types something we can only apply to tests?

src/lib/shopping-fpti/event-handlers.js
zacharyhill: Yeah, that makes more sense. I'll update it.

src/lib/shopping-fpti/event-handlers.js
ukalbhor: hmm yeah ..... the tests need to fail which will be complicated with a `@flow` type

ukalbhor (approved):

PR #157, opened 2022-02-07
DTSHOPSDK-494: Promisify user identification.

yehor-anisimov:
Split get country, and identify user in a smaller chunks of code.
Minor code improvement

### DIFF
--- a/iframes/identity/identity.js
+++ b/iframes/identity/identity.js
@@ -1,54 +1,53 @@
-const fetchVisitorInfo = ({ deviceInfo, country }) => {
-  const encodedDeviceInfo = encodeURIComponent(JSON.stringify(deviceInfo))
+import {fetchVisitorInfo} from './userInfo'
+import {fetchUserCountry} from './userCountry'
+import constants from "../../src/lib/constants";
 
-  const fetchOptions = {
-    method: 'POST',
-    credentials: 'include',
-    mode: 'cors',
-    headers: {
-      'Content-Type': 'application/json'
-    },
-    body: JSON.stringify({
-      query: `{ visitorInfo(
-        country: "${country}",
-        deviceInfo: "${encodedDeviceInfo}"
-      ) }`
-    })
-  }
-
-  return fetch('/targeting/graphql', fetchOptions)
-    .then(res => {
-      if (res.status !== 200) {
-        throw new Error(`targeting responded with statuscode ${res.status}`)
-      }
-
-      return res.json()
-    })
-    .then(json => {
-      if (!json || !json.data || !json.data.visitorInfo) {
-        throw new Error('response missing required fields')
-      }
-
-      return json.data.visitorInfo
-    })
-}
+const {IDENTITY_MESSAGES} = constants;
 
 window.addEventListener('message', async (e) => {
-  if (e.data.type !== 'fetch_identity_request') {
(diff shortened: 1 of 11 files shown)


iframes/identity/identity.js
```diff
@@ -1,54 +1,53 @@
...
+    }, '*')
+
+    window.parent.postMessage({
+      type: IDENTITY_MESSAGES.USER_INFO_REQUEST,
+      payload: {}
+    }, '*')
+  }
+})
+
+window.addEventListener('message', async (e) => {
```
denkt315, line 32: Why do we need to add two listeners? Maybe let's combine in one method and trigger submethods based on the message type.

iframes/identity/userCountry.js
```diff
@@ -0,0 +1,26 @@
+export const fetchUserCountry = () => {
+    const US = 'US';
```
denkt315, line 2: 'US' is used twice for default country. Maybe lets use some constant for this

src/lib/iframe-tools/identity-manager.js
```diff
@@ -59,27 +64,83 @@ export class IdentityManager extends IframeManager {
...
+  //
+  //   debugLogger.log('[identity-manager:fetchIdentity] Fetch identity request.');
+  //   this.iframe.contentWindow.postMessage({
+  //     type: 'fetch_identity_request',
+  //     payload: {
+  //       deviceInfo,
+  //       country
+  //     }
+  //   }, this.url.origin);
+  // }
```
denkt315, line 145: commented-out code.

src/lib/shopping-analytics.js
```diff
@@ -64,7 +64,7 @@ export const shoppingAnalyticsSetup = (config? : Config = {}) => {
   }
   
   debugLogger.log('[shopping-analytics:shoppingAnalyticsSetup] Initialized shopping analytics with the following configuration', config);
-
+  // TODO: promisify setupUserDetails, setupContainer
```
denkt315, line 67: Let's remove todo and create ticket

ukalbhor:
Please keep PR open. I will review it as part of noop.js story (https://engineering.paypalcorp.com/jira/browse/DTSHOPSDK-820) I'm working on this sprint.

src/lib/iframe-tools/identity-manager.js
```diff
@@ -59,27 +64,83 @@ export class IdentityManager extends IframeManager {
     this.completionListener(identity, null);
   }
 
+  fetchCountry = () => {
+    return new Promise((resolve) => {
+      this.addMessageListener(({ data: { type, payload } }) => {
```
luisingho, line 69: Should we unregister this message listener after getting the payload, or is fetchCountry only expected to run once and it wouldn't make much of a difference? Same question for fetchUserInfo

yehor-anisimov: @luisingho This is a good point, I would be happy to do it. Unfortunately, current implementation of message manager doesn't allow it.
I can't include it in current pull request because it is way outside scope of this task.

luisingho (approved):

src/lib/shopping-analytics.js
yehor-anisimov: Comment removed. I'm not sure is it worth separate ticket. Benefits are marginal.

denkt315 (approved):

PR #159, opened 2022-03-03
MP-176: Stop asking country for 50% of the traffic

ukalbhor:


src/lib/iframe-tools/identity-manager.js
```diff
@@ -87,12 +87,28 @@ export class IdentityManager extends IframeManager {
...
+
+    if (shouldReallyCheckCountry) {
+      this.fetchCountry().then((country) => {
+        debugLogger.log('[identity-manager:fetchIdentity] Fetch identity request. Country:', country);
+        this.fetchUserInfo(country);
+      });
+    } else {
+      debugLogger.log('[identity-manager:fetchIdentity] Fetch identity request. Country not fetched.');
+      this.fetchUserInfo('ALL');
+    }
```
mikeAtHoney, line 111: Any unit test additions needed to verify these functional changes?

src/lib/iframe-tools/identity-manager.js
```diff
@@ -87,12 +87,28 @@ export class IdentityManager extends IframeManager {
...
+    } else {
+      shouldReallyCheckCountry = LS.getItem(localStorageKey);
+    }
 
-      debugLogger.log('[identity-manager:fetchIdentity] Fetch identity request. Country:', country);
-      this.fetchUserInfo(country);
-    });
+    debugLogger.log(`[identity-manager:fetchIdentity] Fetching Identity`);
+
+    if (shouldReallyCheckCountry) {
```
mikeAtHoney, line 103: I think we could also remove the need for `shouldReallyCheckCountry` if the suggestion above is followed: https://github.com/paypal/paypal-muse-components/pull/159/files#r819142163.  The if could just check `LS.getItem(localStorageKey)`.  Also `localStorageKey` could use a more specific name like `shouldReallyCheckCountryKey`.

src/lib/iframe-tools/identity-manager.js
```diff
@@ -87,12 +87,28 @@ export class IdentityManager extends IframeManager {
...
+    let shouldReallyCheckCountry;
+    const localStorageKey = 'pp-sdk-shouldReallyCheckCountry';
+    const LS = window.localStorage;
+    if (LS.getItem(localStorageKey) === null) {
+      const randomChance = Math.floor(Math.random() * 100);
+      shouldReallyCheckCountry = randomChance < 50;
+      LS.setItem(localStorageKey, shouldReallyCheckCountry);
+    } else {
+      shouldReallyCheckCountry = LS.getItem(localStorageKey);
+    }
```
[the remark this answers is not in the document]
ukalbhor: I like this suggestion and will adopt it ... but the readability suffers a bit .....  because using `shouldReallyCheckCountry` preserves the context throughout the code block ....

src/lib/iframe-tools/identity-manager.js
ukalbhor: - `shouldCheckCountry` makes the condition more readable.
- There is a `constants.js` which has the keys used with localStorage .... however this code is to start a temporary ramp (currently at 50%) and will be removed in the next sprint once it's at 100%.

zacharyhill (approved):

src/lib/iframe-tools/identity-manager.js
```diff
@@ -87,12 +87,27 @@ export class IdentityManager extends IframeManager {
       return;
     }
 
-    debugLogger.log('[identity-manager:fetchIdentity] Fetch country request.');
-    this.fetchCountry().then((country) => {
+    const localStorageKey = 'pp-sdk-shouldReallyCheckCountry';
+    const LS = window.localStorage;
```
gregjopa, line 91: Please wrap this window.localStorage usage in a try/catch. The `getItem()` and `setItem()` can throw errors if local storage is full or not available.

The belter utility library does have a `isLocalStorageEnabled()` function you can use which safely checks for localStorage support with a try/catch: https://github.com/krakenjs/belter/blob/2f03832059dabc709d296a624e9798a2e8d66f44/src/dom.js#L306

src/lib/iframe-tools/identity-manager.js
yehor-anisimov: I don't think so. Our grand plan to remove country check at all. It is temporary solution, just to be sure we don't have any unexpected results

src/lib/iframe-tools/identity-manager.js
yehor-anisimov, line 91: Can we use sessionStorage, since it has self cleaning mechanism?

src/lib/iframe-tools/identity-manager.js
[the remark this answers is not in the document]
mikeAtHoney: No problem.  The Honey JS coding standards that our previous team followed avoids `let` when possible in favor of immutability but I'm good with what you originally suggested.

src/lib/iframe-tools/identity-manager.js
ukalbhor: I thought about that, but we want the `shouldFetchCountry` decision to always persist for a user so that we have a noiseless picture in our graphs. If one day you fetchCountry and the next you don't, it is inconsistent.

PR #181, opened 2023-05-02
DTSHOPSEOL-1832: [Store-cash] SDK, throw error on merchant website

yehor-anisimov:
Fix null pointer exception.
It happens when merchant use multiple sdks, and some iframe became detached.


src/lib/iframe-tools/iframe-manager.js
```diff
@@ -21,6 +21,10 @@ export class IframeManager {
   };
 
   _onMessage = (e) => {
+    if (!e || !e.source) {
```
ukalbhor, line 24: sugg: can you add a comment here about the issue that caused you to add the condition ?

ukalbhor:
>  when merchant use multiple sdks, and some iframe became detached.

this is when a merchant uses multiple `muse` sdks ? what would a valid use cases for using multiple `muse` be ?

ukalbhor (approved):

denkt315 (approved):

PR #183, opened 2023-07-05
Conditionally add disableSetCookie query param to pptm.js request

spencersablan:
As part of our cookie remediation effort, we need to conditionally add this param so that we could communicate to Stargate to not set non-essential cookies in the response.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/muse-components",
-  "version": "1.3.92",
+  "version": "1.3.91",
   "description": "MUSE component for unified PayPal/Braintree web sdk",
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -68,7 +68,9 @@
       "__WEB__": true,
       "__CORRELATION_ID__": "abc123",
       "__SDK_HOST__": "test.paypal.com",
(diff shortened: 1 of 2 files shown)


src/component.js
```diff
@@ -46,6 +46,11 @@ export function getPptmScriptSrc(paypalDomain : string, mrid : ?string, clientId
   if (getSDKQueryParam('components')) {
     src += `&comp=${ String(getSDKQueryParam('components')) }`;
   }
+
+  // This is a temporary param that will allow us to ramp changes for the cookie remediation effort.
+  if (getDisableSetCookie()) {
+    src += 'disableSetCookie=true';
```
gregjopa, line 52:
```suggestion
    src += '&disableSetCookie=true';
```

spencersablan:
Todo for later: Figure out why tests are failing.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/muse-components",
-  "version": "1.3.91",
+  "version": "1.3.92",
```
gregjopa: The release process should take care of bumping the version. No need to do it in this PR
```suggestion
  "version": "1.3.91",
```

spencersablan: Yeah, I just fixed this 😆  This was leftover from when I was trying to fix the tests.

PR #186, opened 2023-08-25
Add disable-set-cookie header to /targeting/graphql

spencersablan:
Need to add this flag so that the cookie experiment works with `/targeting/graphql`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -22,6 +22,7 @@ module.exports = {
   ],
   'plugins': [
     'babel-plugin-transform-es2015-modules-commonjs',
+    '@babel/plugin-transform-modules-commonjs',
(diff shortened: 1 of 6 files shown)


babel.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
   ],
   'plugins': [
     'babel-plugin-transform-es2015-modules-commonjs',
```
djake, line 24: can this be removed now that you added `@babel/plugin-transform-modules-commonjs`?

ukalbhor (approved):

babel.config.js
yehor-anisimov: Agree, looks like unnecessary plugin.

djake (approved):

yehor-anisimov (approved):

